PR #10, opened 2026-06-11
Gate New Asta Features

AmberRose2:
Added a new permission `enroll:asta_integration` that gates the new "explore with asta" features so that only users that have the correct permission can see them.

Tested and validated that without the permission nothing was visible, and with it I was able to see the new faetures.

I also switched to point to prod Asta since changes are now deployed there and tested to make sure everything looked good.


AmberRose2:
With  the `enroll:asta_integration` perm:
Screenshot 2026-06-15 at 10 29 47 AM

Without the `enroll:asta_integration` perm:
Screenshot 2026-06-15 at 10 34 13 AM

regan-huff (approved):

PR #5, opened 2026-06-09
Improvements to static HTML report

rodneykinney:
- Cut down on the size by stripping unnecessary information, splitting figures into standalone files.
 - First build with the new repo structure. 
 - Add `RELEASE.md` for pypi project home page
 - Bump version to 0.2.1

PR #11, opened 2026-06-12
Public pypi documentation

rodneykinney:
Create a separate file for the pypi project home page


rodneykinney:
Folding into https://github.com/allenai/asta-autodiscovery/pull/5

PR #13, opened 2026-06-16
Show experiment context in Continue with Asta modal

MLatzke:
image

- Add ExperimentContextSummary component (Belief Shift, Hypothesis, Analysis), shared between the detail panel and modal so they stay in sync
- Render experiment context in a left panel of the modal with the experiment ID as a title; thread surprisalWidth through for matching surprisal coloring
- Two-column layout (2:1 context:prompt) with a full-bleed divider and a darker Explore panel for clearer separation
- On small screens, stack the panels, fill near-full height within dialog margins, and show a scroll-affordance shadow on the context column

This handles:
https://github.com/allenai/nora-issues/issues/2938
https://github.com/allenai/nora-issues/issues/2924

### DIFF
--- a/ui/src/app/runs/components/ContinueWithAstaModal.tsx
+++ b/ui/src/app/runs/components/ContinueWithAstaModal.tsx
@@ -7,9 +7,11 @@ import {
     Dialog,
     DialogContent,
     IconButton,
+    Stack,
     TextField,
     Typography,
     styled,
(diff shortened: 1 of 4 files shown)


AmberRose2 (approved):
Yay looks great - I tested all sorts of widths and on mobile and all look good! :)

PR #20, opened 2026-06-17
Parameterize Auth0 Management API domain via AUTH0_MGMT_DOMAIN

dirkraft:
## Problem

The scheduled `autodiscovery-send-emails-prod` Cloud Run job is failing on **every** user with:

```
Failed to get Auth0 management token: [SSL: CERTIFICATE_VERIFY_FAILED]
certificate is not valid for 'YOUR_TENANT.us.auth0.com'
```

The Auth0 tenant domain for the Management API was hardcoded to a placeholder constant (`AUTH0_MGMT_DOMAIN = "YOUR_TENANT.us.auth0.com"`), so the job tries to fetch a management token from a host whose cert doesn't match — every email lookup fails and the job exits non-zero (0 emails sent).

This was a regression: the domain used to be read from the environment but was changed to a hardcoded constant.

## Fix

Read the domain from the `AUTH0_MGMT_DOMAIN` environment variable, matching how `AUTH0_MGMT_CLIENT_ID` / `AUTH0_MGMT_CLIENT_SECRET` are already supplied. No ai2-specific default remains in the public repo — the value is injected by deployment config.

- `auth0.py`: read `AUTH0_MGMT_DOMAIN` via a small `_require_env()` helper (raises `Auth0Error` if unset); drop the hardcoded module constant.
- `.env.example`: add `AUTH0_MGMT_DOMAIN=`.
- `send_completion_emails.py` / `export_users_csv.py`: document the new required env var.

## Deploy notes

`AUTH0_MGMT_DOMAIN` must be set on the email/cleanup Cloud Run jobs (the CI image auto-update only swaps `--image`, not env vars). Tracked on the deployment side.


dirkraft:
Pretty much only need to lookup user emails. So the AD web stack and AD jobs are completely unaffected.

AmberRose2 (approved):
Oop - good catch 🫡

PR #21, opened 2026-06-17
Make start.sh the default api ENTRYPOINT for Skiff2 (Cloud Run)

gas2own:
## Why
Companion to [`asta-autodiscovery-private#378`](https://github.com/allenai/asta-autodiscovery-private/pull/378) (Skiff2 bootstrap). Skiff1 is being retired.

Under Skiff1 the api container's command was overridden to `/api/start.sh` in `.skiff/webapp.jsonnet` (`command: ['/api/start.sh']`), while this image's `ENTRYPOINT` defaulted to `/api/dev.sh`. **Skiff2 / Cloud Run cannot override the container command**, so the production entrypoint has to be the image default — otherwise the api would run `dev.sh` (gunicorn with `--reload` and non-structured logs) in production.

## What
- **`api/Dockerfile`**: `ENTRYPOINT` → `/api/start.sh`. `start.sh` is the prod gunicorn invocation (no `--reload`, structured JSON access logs); `dev.sh` is identical except for `--reload` and plain logging.
- **`docker-compose.yaml`**: add `entrypoint: ["/api/dev.sh"]` to the `api` service. The compose api service bind-mounts `./api:/api` and had **no** command override, so it relied on the image default for hot reload. This explicit override preserves the exact local-dev behavior (`--reload`) after the Dockerfile default flips.

Net effect: production (Skiff1 today, Skiff2 after cutover) and local dev both keep their current behavior — the change just relocates *where* each entrypoint is selected.

## Verification
- `start.sh` / `dev.sh` both exist in `api/`; diff between them is only `--reload` + access-log format.
- Confirmed `docker-compose.yaml` api service has no other `command`/`entrypoint`.


dirkraft:
Weird bleed out from internal infra to public repo. Tolerable but not ideal.

radamson (approved):
👍

PR #22, opened 2026-06-18
Require SMTP_SERVER and SENDER_EMAIL env vars for the email job

dirkraft:
## Problem

After the Auth0 fix (#20), the send-completion-emails job got past user lookup but then failed at the send step:

```
ERROR - Failed to send email for <user>/<run>: [Errno 8] nodename nor servname provided, or not known
```

`email.py` hardcoded ai2-specific defaults (`smtp.example.com`, `no-reply@allenai.org`) and the job never overrode the server, so every send tried to resolve the placeholder host and failed (DNS). Same scrub-placeholder class of bug as the Auth0 domain.

## Fix

Remove both defaults from the public repo and require `SMTP_SERVER` and `SENDER_EMAIL` from the environment (raising `EmailError` if unset), mirroring the `AUTH0_MGMT_*` vars. `SMTP_PORT` stays optional (defaults to 25). The real values are injected by deployment config.

- `email.py`: drop `DEFAULT_SMTP_SERVER` / `DEFAULT_SENDER_EMAIL`; resolve `SMTP_SERVER` / `SENDER_EMAIL` / `SMTP_PORT` from the environment via a `_require_env()` helper.
- `__init__.py`: stop exporting the removed defaults; export `EmailError`.
- `.env.example`: add `SMTP_SERVER` / `SENDER_EMAIL`.
- `send_completion_emails.py`: document the required vars.

## Import safety

The required-var checks are **lazy** — they run inside `send_email()`, not at import time. Importing the `autodiscovery_jobs` package (as the API app and the dataset-cleanup job do) needs no email config; only the email job, the sole caller of `send_email()`, does. Verified:

```
$ env -u SMTP_SERVER -u SENDER_EMAIL python -c "import autodiscovery_jobs; from autodiscovery_jobs import JobConfig, send_email"
import OK — no env vars needed
# send_email() raises only when actually called without config
```

## Deploy notes

The email Cloud Run jobs must set `SMTP_SERVER` and `SENDER_EMAIL` (in addition to `AUTH0_MGMT_DOMAIN`). The CI image auto-update only swaps `--image`, not env vars. Tracked on the deployment side.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -30,6 +30,11 @@ AUTH0_MGMT_DOMAIN=
 AUTH0_MGMT_CLIENT_ID=
 AUTH0_MGMT_CLIENT_SECRET=
 
+# Email job: mail server and from-address (required to send; port defaults to 25)
+SMTP_SERVER=
+SENDER_EMAIL=
+# SMTP_PORT=25
+
 # Optional: Required Auth0 permission for authenticated users to access the API.
 # If not set or empty, any authenticated user can access the system.
 # Example: AUTH0_REQUIRED_PERMISSION=enroll:autodiscovery_v0
--- a/packages/autodiscovery_jobs/src/autodiscovery_jobs/__init__.py
+++ b/packages/autodiscovery_jobs/src/autodiscovery_jobs/__init__.py
@@ -90,9 +90,8 @@ from .auth0 import (
 
 # Email sending
 from .email import (
+    EmailError,
     send_email,
-    DEFAULT_SENDER_EMAIL,
-    DEFAULT_SMTP_SERVER,
 )
 
 __version__ = "0.2.2"
@@ -154,7 +153,6 @@ __all__ = [
     # Auth0 functions
     "get_user",
     # Email functions
+    "EmailError",
     "send_email",
-    "DEFAULT_SENDER_EMAIL",
-    "DEFAULT_SMTP_SERVER",
 ]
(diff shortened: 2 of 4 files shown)


dirkraft:
Set all the required vars and was actually able to trigger some emails finally

```
➜  public git:(fix/parameterize-smtp-server) uv run python scripts/send_completion_emails.py --max-age-hours 48
2026-06-18 10:54:00,819 - INFO - ============================================================
2026-06-18 10:54:00,819 - INFO - Send Completion Emails Script
2026-06-18 10:54:00,819 - INFO - Bucket: autodiscovery
2026-06-18 10:54:00,819 - INFO - Max age: 48 hours
2026-06-18 10:54:00,819 - INFO - Dry run: False
2026-06-18 10:54:00,819 - INFO - User filter: all users
2026-06-18 10:54:00,819 - INFO - ============================================================
2026-06-18 10:54:00,819 - INFO - Scanning for runs completed after 2026-06-16T17:54:00.819971+00:00
2026-06-18 10:54:03,842 - INFO - Found 1985 users to scan (skipped 4 non-Auth0 users)
2026-06-18 10:55:09,913 - INFO - Email sent successfully to ...
2026-06-18 10:55:10,321 - INFO - Sent completion email for ...
2026-06-18 10:55:20,548 - INFO - Email sent successfully to ...
...
```

AmberRose2:
Ahhh I see - so this is what was causing strange behavior with emails -https://github.com/allenai/nora-issues/issues/2927

AmberRose2 (approved):
I don't think there's really a use case where a dev would need to set these env variables for testing locally (other than you right now) but still might be nice to mention it in the env example [here](https://github.com/allenai/asta-autodiscovery/blob/main/.env.example)

Which reminds me I need to update that as well lol

dirkraft:
'tis 
Screenshot 2026-06-18 at 3 01 26 PM

PR #23, opened 2026-06-23
use context service instead of writing to bucket directly

goblinbrooke:
Part of: https://github.com/allenai/nora-issues/issues/2937

Context service lives here: https://github.com/allenai/asta-context-service

In order to allow auto-delete after a certain amount of time (30 days) to mimic DV, we want to use the context service instead of directly uploading to the GCS bucket, so we can add tags and create an auto-delete job. This switches auto-discovery over to that service. 

Next steps: 
In the context service, creating the job that deletes with these tags and ensuring it runs daily


api/runs/runs_api.py
```diff
@@ -1586,17 +1587,24 @@ def dig_deeper_with_asta(userid: str, runid: str, experiment_id: str):  # pyrigh
...
         try:
-            asta_gcs.save_manifest_json(user_uuid, thread_id, manifest.model_dump())
+            manifest_gcs_uri = asta_context_client.upload_json_artifact(
+                owner_id=user_uuid,
+                prefix=thread_id,
+                filename="manifest.json",
+                content=manifest.model_dump(),
+                artifact_type="manifest",
+                source="autodiscovery",
+            )
```
cecile-nguyen, line 1604: do you want to pass the delete date tag here to?

cecile-nguyen (approved):

api/runs/runs_api.py
goblinbrooke: Yes good call!

PR #25, opened 2026-06-24
Make it so the pop-up isn't blocked

AmberRose2:
PR addressing this report from the bug bash: https://github.com/allenai/nora-issues/issues/2922

Previously the Asta tab was opened after an async API call, which browsers flag as a pop-up. Now a tab is opened synchronously on click (within the user gesture), then navigated to the Asta URL once the API resolves.

This creates a bit of waiting time so I created a basic loading screen — rather than a jarring blank white tab, the interim tab now shows the Asta logo on a dark teal background with a subtle pulse animation, matching the app's existing branding. Matt can make it prettier when he comes back if he would like haha.

https://github.com/user-attachments/assets/64cbe2fb-9e63-4b29-8a8b-78464e3ca09e


pauls-ai2 (approved):
Let's give it a shot!

PR #26, opened 2026-06-25
set dev to hit localhost

AmberRose2:
Temporarily have dev AD send requests to localhost so I can test some changes in the nora code locally before opening a PR


dirkraft:
do you still need this?

AmberRose2:
> do you still need this?

I don't think so - thank you!

PR #28, opened 2026-06-30
fix public sample dataset links

dirkraft:
The bucket isn't publicly accessible and the name of it isn't particularly sensitive so these can stay where they are.

Long term I think we're likely to completely change how input files are widgetized and specified to agentic research systems, so this is really just keeping the current functionality intact with as little change as possible.


dirkraft:
selected kidney dataset

Screenshot 2026-06-30 at 12 48 54 PM

AmberRose2 (approved):
Mucho gracias!

PR #29, opened 2026-07-01
Swappable auth providers (auth0 / password_file / none)

dirkraft:
High level context: This is one of a series of transformations to make AutoDiscovery more portable and adaptable into other non-Ai2 environments. #34 

----

## Summary

Makes authentication pluggable across the Flask API and the Next.js UI. One `AuthProvider` interface, implemented three times, selected by the `AUTH_PROVIDER` env var and discovered by the UI at runtime via a new `GET /api/auth/config` — so a **single build serves all three modes**.

- **`auth0`** (default) — existing Auth0 behavior, unchanged.
- **`password_file`** — a `user:password` file (bcrypt) administered by a Python CLI, mounted read-only into the API; re-read on **every** request so new/changed/disabled users take effect with no restart.
- **`none`** — "desktop mode": every request is a fixed `local` user with all permissions.

Full design + rationale: `docs/design/auth-providers.md`.

## Backend (`api/`)

- `utils/auth.py` → `utils/auth/` package. The `requires_auth` / `requires_enrollment` / `optional_enrollment` decorators **keep their public API and the `request.user` dict shape**, delegating to the active provider — so **none of the ~23 endpoint call sites changed**.
- `Auth0Provider` lifts the JWKS/JWT/`/userinfo` logic over verbatim (behavior-identical).
- `PasswordFileProvider`: `POST /api/auth/login` verifies credentials once and mints a short-lived HS256 session token; each request re-reads the file for authoritative user/permission checks.
- `NoneProvider`: fixed local user, never raises.
- `scripts/auth_admin.py`: `useradd/passwd/usermod/disable/enable/userdel/list`, sharing the store/bcrypt module with the provider. Adds `bcrypt` to requirements.
- New `auth` blueprint: `GET /api/auth/config`, `POST /api/auth/login`.

## Frontend (`ui/`)

- Rewrote the auth context into a provider-agnostic context + runtime selector + three implementations. **Kept the `useAuth0()` / `Auth0Provider` export names** (added `useAuth` alias), so the ~20 consumers and `ClientProviders` are untouched.
- Added centralized `permissions` / `hasPermission()`, replacing the two ad-hoc client-side JWT decodes.
- `BaseApi` reads tokens from a new `authBridge` singleton instead of importing Auth0.
- `LoginDialog` for password mode; `loginWithRedirect` repurposed as the generic interactive-login trigger (redirect / open form / no-op); `AuthButton` hides under `none`.

## Testing

- Backend: verified `/api/auth/config`, `/api/auth/login` (good/bad creds), and `/api/user/me` for all three providers in **fresh processes**; confirmed the password-file provider picks up disable + permission changes live.
- Frontend: `yarn ts-check` and `yarn lint` pass clean.


dirkraft:
run using AUTH_PROVIDER=none

Screenshot 2026-07-07 at 2 17 28 PM

api/utils/auth/permissions.py
```diff
@@ -0,0 +1,13 @@
+from enum import Enum
+
+
+class PermissionType(Enum):
```
regan-huff, line 4: I'm not sure I see the way around it, but this is ai2-flavored

regan-huff (approved):
I think this is a good workable approach to maintaining the auth structure ai2 autodiscovery needs but allowing other users to adapt or not auth. I think the single/local user default to "local" is fine. It's impossible to know what everyone might want; this enables our organization's case as well as the simplest local user case and it's a reasonable solution.

api/utils/auth/permissions.py
dirkraft: If i get through everything else and we don't have bigger fish to fry, i may come back around to this. At the very least I can make the organization of the source code draw a box around asta-specific parts by moving those into an asta subpackage.

dirkraft:
here's AUTH_PROVIDER=password_file

```
uv run api/scripts/auth_admin.py useradd jasond --email jasond@allenai.org --password blah
```

image

PR #30, opened 2026-07-01
Swappable AutoDiscovery job backend: GCP Cloud Run + local Docker

dirkraft:
A step closer to minimizing external dependencies and making AD adaptable to other environments. Epic described in #34 

Per this branch...
* no dependency on auth0
* this PR: no dependency cloud run job for AutoDiscovery process - JOB_BACKEND=docker (new default)
* still dependency on modal code sandbox (next)
* still dependency on gcs bucket as the shared disk (nextnext)

----

## Context

The webstack launches AutoDiscovery (AD) jobs as **GCP Cloud Run job executions**. This PR extracts the AD job *launcher* behind a `JobBackend` interface so it becomes swappable between:

- **`gcp`** — Cloud Run executions (current behavior, still the default)
- **`docker`** — local Docker containers (for local dev / portability), launched by the API via a mounted `docker.sock`

Only the job *process launcher* changes. The AD job itself continues to use **Modal** for code-execution sandboxes, unchanged — and Modal's per-user/per-job read-only dataset isolation (`asta_sandbox.CloudShare`, driven by `--bucket_path`) is untouched.

## How it works

- The CLI args handed to the AD image are backend-agnostic, so they're built once in `build_job_args()` and reused by both backends.
- Backend is chosen per deployment via **`JOB_BACKEND`** (`gcp` | `docker`), read in `JobConfig.from_env` and plumbed through docker-compose as a passthrough so the same local stack can target either without editing compose.
- **GCS for the Docker job:** there's no Cloud Run FUSE volume locally, so the AD job container **self-mounts** the real bucket at `/mnt/gcs` via `gcsfuse` in its entrypoint — gated on `GCSFUSE_BUCKET` so the Cloud Run path (where the platform provides the mount) is unaffected. No host-daemon plugin required.

## Changes

**New `autodiscovery_jobs/backends/` subpackage**
- `base.py` — `JobBackend` ABC + `build_job_args()`
- `gcp.py` — `CloudRunBackend` (existing Cloud Run logic moved here)
- `docker.py` — `DockerBackend` (docker-py over `docker.sock`); forwards Modal/OpenAI/Vertex/GCP env, bind-mounts the GCP key, grants `/dev/fuse` + `SYS_ADMIN`
- `__init__.py` — `get_backend(config)` factory

**Wiring**
- `cloudrun.py` → thin compat shim re-exporting the functional API
- `config.py` → `backend` (`JOB_BACKEND`) + `job_image` (`AUTODISCOVERY_IMAGE`)
- `manager.py` → delegates execution to `self.backend`; **fixes a latent `cloudrun.cancel_execution` bug** in `soft_delete_job`
- `run_details.refresh_run_status` → uses the backend factory
- `pyproject.toml` / `uv.lock` → add `docker>=7.0`

**Job image + compose**
- `packages/autodiscovery/Dockerfile` + `docker-entrypoint.sh` → gcsfuse self-mount gated on `GCSFUSE_BUCKET`
- `docker-compose.yaml` → `docker.sock` mount, `JOB_BACKEND`/`AUTODISCOVERY_IMAGE`/`GCP_KEY_HOST_PATH` passthrough, build-only `autodiscovery` service (profile `jobs`); `.env.example` documents new vars

## Testing

- **67 tests pass** (existing suite + new `test_backends.py` covering the factory, `build_job_args` parity, and the Docker backend). Ruff clean on authored files.
- Verified in an isolated Python 3.13 venv — note the local workspace `.venv` is a broken Python 3.14 (matplotlib won't build), unrelated to this change.
- **Not run (needs live stack):** end-to-end Docker-backend submit → SUCCEEDED → outputs in GCS. To try: set `JOB_BACKEND=docker` and `GCP_KEY_HOST_PATH` in `.env`, `docker compose build autodiscovery`, then `docker compose up`.


dirkraft:
with AUTH_PROVIDER=none and JOB_BACKEND=docker

Screenshot 2026-07-23 at 10 23 31 AM

```
➜  asta-autodiscovery2 git:(swappable-job-backend) docker ps -a
CONTAINER ID   IMAGE                       COMMAND                  CREATED          STATUS                        PORTS                                         NAMES
00358fe456f8   autodiscovery:dev           "/usr/local/bin/dock…"   12 minutes ago   Exited (0) 4 minutes ago                                                    autodiscovery-job-dev-a3378d59
...

➜  asta-autodiscovery2 git:(swappable-job-backend) 2>&1 docker logs autodiscovery-job-dev-a3378d59 | tail -10
[Exploration] Time elapsed: 413.89 seconds.

[CSV] MCTS nodes (n=4; skipping root) saved to /mnt/gcs/users/local/jobs/68fdd533-a0c9-48f9-aff5-2d2b0032be50/output/mcts_nodes.csv.


DELETED WORKING DIRECTORY: work

RUN FINISHED!

LOGS: /mnt/gcs/users/local/jobs/68fdd533-a0c9-48f9-aff5-2d2b0032be50/output
```

.env.example
```diff
@@ -5,15 +5,33 @@
 
 GOOGLE_ACCESS_KEY_ID=GOOG1E...
 GOOGLE_ACCESS_KEY_SECRET=...
-GOOGLE_APPLICATION_CREDENTIALS=./secrets/gcp-key.json  # Need to make a secure directory and place your GCP key file here
+# Path to your GCP key file. Use an ABSOLUTE path when using the docker job backend
+# (the default): the API bind-mounts this file into each job container, and the
+# host daemon needs an absolute source path. e.g. /Users/you/asta/secrets/gcp-key.json
+GOOGLE_APPLICATION_CREDENTIALS=./secrets/gcp-key.json
```
AmberRose2: Since this is just an example why not but the full path here too? I.e.
`GOOGLE_APPLICATION_CREDENTIALS= /Users/you/asta/secrets/gcp-key.json`

packages/autodiscovery_jobs/src/autodiscovery_jobs/backends/docker.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+class DockerBackend(JobBackend):
+    """:class:`JobBackend` that launches jobs as local Docker containers.
+
+    The ``execution_id`` returned by :meth:`run_job` is the container name, and
+    is passed back to the other methods to inspect / stop / read logs.
+    """
+
+    def run_job(self, userid: str, jobid: str, **kwargs) -> str:
```
AmberRose2, line 72: Now that we are moving to docker job containers should we pass in memory / cpu limits to protect against a case where a job is eating up all the resources and starving the host?

AmberRose2 (commented):
Some small comments mostly as things that might be helpful down the line. Lots of lines of code so did a more general pass / sanity check - but looks good. 

The screenshot you posted is of an end-to-end Docker-backend submit yes? If that worked with no error, combined with the code here I feel good about merging!

.env.example
dirkraft: indeed

packages/autodiscovery_jobs/src/autodiscovery_jobs/backends/docker.py
dirkraft: For us folks on mac using Docker Desktop, you already choose limits over all docker activity, protecting your machine a little bit from whatever is going on in docker. I believe linux docker installations have an equivalent, but I bet this depends on the exact installation and linux you're using.

For deployed infrastructure, they all have their own mechanics in their own ways. GCP cloud run jobs you set those limits on the job definition (not per invocation). So that provisioning of the job definition itself happens before you're ever in this code launching an instance of an AD job; and then I wondered can you override mem limits per invocation of a cloud run job? Answer seems to be no, but other backends out there might allow that. Even docker, depending on the exact details of your docker installation have different capabilities in terms of limiting resource usage.

I think it's still a good point, but i think its hard to provide those options without doing some other things to make it reasonably useful. e.g. i set memory limit per AD job, but then i start n concurrent jobs and hit the docker engine limits anyways. If someone is at the point of customizing AD for their needs, then I'm gonna bet they'll be customizing the code anyways, at which point, they change the code with the behavior they want.

AmberRose2 (approved):

PR #31, opened 2026-07-01
perf(credits): reuse one GCS client, drop redundant existence checks

dirkraft:
## Problem

`GET /api/user/me/credits` takes **~8s** on every load (for jason's user id). `get_user_credits()` walks *every* job for a user in GCS, and each of the small GCS helpers (`get_metadata`, `count_experiment_results`, `get_run_details`, `get_user_profile`, …) constructed a **brand-new `storage.Client`**.

Instrumenting a single credits request for a real 25-job account:

```
storage.Client() constructed 88 times for one credits request
storage.Client() construction measured at ~66ms each  →  ~5.8s of the 8s is pure client setup
```

On top of that, `get_metadata` / `get_job_args` did a separate existence-check round-trip (`job_exists()` list / `blob.exists()` HEAD) *before* every download.

## Changes

1. **Reuse a single GCS client.** New `autodiscovery_jobs.client.get_storage_client()` returns a process-wide cached `storage.Client` per `project_id` (thread-safe, double-checked lock). The `gcs`, `run_details`, `user_profile`, and `email_state` helpers now obtain the client from it instead of each constructing its own. Reusing a long-lived client is the pattern Google's client libraries recommend (credentials refresh transparently).

2. **Drop redundant existence checks.** `get_metadata` and `get_job_args` now download directly and handle `NotFound`, instead of a pre-check round-trip. `get_metadata` **preserves its exception contract** (`JobNotFoundError` vs `GCSError`) by falling back to `job_exists()` only on a miss — so the common "metadata exists" path is a single round-trip while callers that distinguish the two exceptions (e.g. `runs_api.py`) are unaffected.

3. **Test fixture.** `mock_storage_client` now patches the new construction site (`autodiscovery_jobs.client.storage.Client`) and clears the per-process cache between tests.

## Results

Measured with the real `get_user_credits()` against a 25-job account in `gs://ai2-autodiscovery`:

| | before | after |
|---|---|---|
| latency | ~8.0s | **~0.4s** (0.66s cold) |
| `storage.Client()` per request | 88 | **1** |
| result | `granted=1000 consumed=420 pending=0 available=580` | **identical** |

## Testing

- `pytest packages/autodiscovery_jobs/tests/` — all pass (the one unrelated `test_manager_default_config` failure is a pre-existing env-var leak, fails identically on `main`).
- `ruff check` — no new findings (net -1 vs baseline).
- End-to-end `get_user_credits()` run against real data, result verified identical.

### DIFF
--- a/packages/autodiscovery_jobs/src/autodiscovery_jobs/client.py
+++ b/packages/autodiscovery_jobs/src/autodiscovery_jobs/client.py
@@ -1,0 +1,52 @@
+"""Shared, cached Google Cloud Storage client.
+
+Constructing a :class:`google.cloud.storage.Client` is surprisingly expensive:
+each instance re-runs credential/session setup (tens of milliseconds apiece).
+The credit and run endpoints call many small GCS helpers per request, and each
+helper used to build its own client, so a single request could construct dozens
+of clients and spend seconds purely on client setup.
+
+A ``storage.Client`` is safe to share across threads for the read/write blob
+operations used here, and the Google client libraries recommend reusing a
+single long-lived client (it transparently refreshes credentials). We therefore
+cache one client per ``project_id`` for the lifetime of the process.
+"""
+
+from __future__ import annotations
+
+import threading
+
+from google.cloud import storage
+
+from .config import JobConfig
+
+_clients: dict[str | None, storage.Client] = {}
+_lock = threading.Lock()
+
+
+def get_storage_client(config: JobConfig | None = None) -> storage.Client:
+    """Return a process-wide cached ``storage.Client`` for the config's project.
+
+    Reuses a single client per ``project_id`` to avoid repeated credential and
+    HTTP-session setup. Thread-safe.
+
+    Args:
+        config: Job configuration (uses default if None).
+
+    Returns:
+        A shared ``storage.Client`` scoped to ``config.project_id``.
+    """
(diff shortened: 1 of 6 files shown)


packages/autodiscovery_jobs/src/autodiscovery_jobs/client.py
```diff
@@ -0,0 +1,52 @@
...
+
+from google.cloud import storage
+
+from .config import JobConfig
+
+_clients: dict[str | None, storage.Client] = {}
+_lock = threading.Lock()
+
+
+def get_storage_client(config: JobConfig | None = None) -> storage.Client:
```
dirkraft: if this only needs project_id, isn't it better to only accept that arg specifically?

dirkraft: Good call — done in 2739ecd. `get_storage_client` now takes `project_id` directly (it only ever used `config.project_id`), which also drops the `JobConfig` import/coupling from `client.py`. All call sites already resolve `config` first, so they pass `config.project_id`. Re-verified: 47 tests pass, ~0.4s, still 1 client per request.

pauls-ai2 (approved):

dirkraft:
lcoal stack definitely _feels_ faster. let's see...

PR #33, opened 2026-07-06
Pass AD Link to Asta

AmberRose2:
This is the work Autodiscovery side for https://github.com/allenai/nora-issues/issues/2929

 - `api/utils/asta_client.py`: create_thread() takes an optional autodiscovery_link param and sends it in the PUT request body as {"Autodiscovery_link": "..."} when set.
  - `api/runs/runs_api.py`: builds the link as `{AUTODISCOVERY_BASE_URL}/runs/{runid}?exp={target_node.creation_idx - 1}` (matching the same exp query param convention the frontend already uses for deep-linking to a specific experiment) and passes it into create_thread.
  - Confirmed via logs that the correct link is constructed and reaches the request body.

This PR also includes a couple changes to how we pass user information (apologies I probably should have done this in two PR's - but one felt easier in the moment lol). This was part of my investigation work on https://github.com/allenai/nora-issues/issues/2935 This might resolve the issue- or not, i was struggling to repro that and am still unsure exactly what caused it, though this was my best guess.

  **Before:** 
One attempt to GET {auth0_domain}/userinfo. If it failed, it logged a warning and silently fell back to email = auth0_user_id, with name/nickname left
  empty — then proceeded anyway to log in with Asta using that fake/partial data.

  **After:**
  - Retries the userinfo fetch up to 2 times on failure instead of giving up immediately.
  - If both attempts fail, it no longer proceeds with fake data — it returns a 502 "Failed to fetch user profile from Auth0" and stops.
   
  **Why:** Asta's login endpoint falls back to matching users by email when auth0_user_id doesn't match an existing record (e.g. the user authenticated via a different Auth0 connection than last time). So sending a missing or wrong email — as the old fallback did — risked Asta creating a duplicate account instead of recognizing the returning user. Failing loudly is safer than silently sending bad data that corrupts account matching.

_note:_
I will open a Partner PR in the nora repo that handles this new param


api/runs/runs_api.py
```diff
@@ -1549,26 +1553,42 @@ def dig_deeper_with_asta(userid: str, runid: str, experiment_id: str):  # pyrigh
...
-                )
-                info.raise_for_status()
-                u = info.json()
-                email = u.get("email", "")
-                name = u.get("name", email)
-                nickname = u.get("nickname", email)
-            except Exception as e:
-                current_app.logger.warning("Could not fetch Auth0 userinfo: %s", e)
-                email = auth0_user_id
+            for attempt in range(2):
```
pauls-ai2, line 1568: This might be overkill, but there is a good library to handle retries, specifically around which errors to retry on. Totally up to you whether it's worth bringing in.

https://scrapeops.io/python-web-scraping-playbook/python-requests-retry-failed-requests/#:~:text=Here%20is%20the%20minimal%20example%3A

api/utils/asta_client.py
```diff
@@ -44,22 +44,31 @@ def login_or_create_user(
...
+        autodiscovery_link: URL back to the source AutoDiscovery experiment,
+            surfaced by Asta so the user can navigate back
 
     Returns:
         Thread key string (use as thread_id for subsequent calls)
 
     Raises:
         requests.HTTPError: If the thread creation call fails
     """
+    body = {"Autodiscovery_link": autodiscovery_link} if autodiscovery_link else None
```
pauls-ai2: Capital "A"?

pauls-ai2: Do you want an empty body, or maybe `{}`?

api/utils/asta_client.py
```diff
@@ -44,22 +44,31 @@ def login_or_create_user(
...
+    profile: str = "dv-a2a-only",
+    autodiscovery_link: str | None = None,
+) -> str:
     """Create a new Asta thread and return the thread key.
 
     Args:
         auth_token: User's bearer token for Asta authentication
         profile: Handler profile to bind to the thread
+        autodiscovery_link: URL back to the source AutoDiscovery experiment,
+            surfaced by Asta so the user can navigate back
```
pauls-ai2, line 58: For my info, this is needed so the link routes to the same env as where the thread was created, correct? That way Asta doesn't need to know what the base URL is?

pauls-ai2 (approved):
Looks good to me! Threw some notes, but totally optional.

api/utils/asta_client.py
AmberRose2: Sorry I'm not sure what you are saying/ asking here! The URL points back to which experiment opened the Asta run from AutoDiscovery (and yes it will save accordingly depending on what AD env the request comes from). BASE_URL refers to the Asta API.

pauls-ai2: An alternative implementation would have been to save the component pieces of the URL as data, then reconstruct the URL on the Asta UI side. For instance, instead of saving `autodiscovery_link`, you could have `run_id`, which you'd then turn into a URL later. By setting the full URL here, it means anything that needs the URL doesn't have to know how to construct a valid URL, but that comes at the cost of the API needing to know what its own URL is. You could potentially pull the API call's URL from the request, then parse that to construct a run's URL.

tl;dr, there are many ways to implement this, and I wanted to make sure I understood why you chose the direction you did; makes it easier to review.

PR #36, opened 2026-07-08
Add configuration reference doc (environment variables)

dirkraft:
## What

Adds `docs/configuration.md`, a single reference for every environment variable the application
reads, and links it in the mkdocs nav (**Home → Configuration**).

## Contents

Variables are grouped by concern, each in a table with **Required**, **Default**, and
**Description** columns:

- Core service (logging)
- Authentication
- Google Cloud & storage
- Job execution
- Modal (code-execution sandbox)
- LLM providers + retry/backoff
- Maintenance jobs (email & user lookup)
- Integration handoff
- Frontend (UI, including `NEXT_PUBLIC_*`) and test/CI variables

Defaults are documented as the code actually uses them, so the page stays accurate against the
current behavior.

## Notes

- This is the initial version. Feature branches that introduce new configurable behavior (e.g.
  the swappable job backend) should extend this doc with their variables.
- No behavior changes — docs only.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,0 +1,150 @@
+# Configuration
+
+AutoDiscovery is configured through **environment variables**. This page documents every
+variable the application reads, grouped by the concern it configures, along with whether it is
+required and its default.
+
+## How configuration is supplied
+
+- **Local development** (`docker compose`): variables are read from a `.env` file at the repo
+  root. Copy [`.env.example`](https://github.com/allenai/asta-autodiscovery/blob/main/.env.example)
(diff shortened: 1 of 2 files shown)


AmberRose2 (approved):
Looks good to me- curious what differentiates between "Required" or not? Like are required variables only the ones needed to spin up a MVP version of DV? 

i.e. The bare minimum to send a AD query and get a result?

dirkraft:
required _should_ mean required to run the application. i don't think that is true due to our inconsistent env var handling, but it will become more true as i work to make this repo more rational and trivial to run outside of Ai2

PR #37, opened 2026-07-08
Change name of Param to be more Generic

AmberRose2:
No need to change internal name - jut when making the call.


pauls-ai2 (approved):
Perfect!

PR #39, opened 2026-07-09
Remove the unused admin package (jobs API + console) and the unused enrollment gate

dirkraft:
## Summary

Removes dead/unauthenticated surface area and the unused enrollment gate. The net diff is almost entirely deletions (~700 lines removed).

### 1. Remove the admin package

`api/static/admin.html` was a standalone operator console served **unauthenticated** at `/api/admin/`, and it drove the equally-unauthenticated `/api/admin/jobs/*` endpoints (`api/admin/jobs_api.py`), which took `userid` as a client parameter — anyone who could reach them could act on any user's jobs.

Those HTTP endpoints turned out to be **only** ever called by that admin console. The webapp manages jobs through `/api/runs/*`, which uses the `autodiscovery_jobs` `JobManager` **in-process** — not this blueprint. With no remaining callers, the entire `admin/` package (jobs API, console, blueprint registration) is removed. `/api/admin*` now 404s.

### 2. Collapse `requires_enrollment` and remove the unused enrollment gate

- `requires_enrollment` is folded into `requires_auth`. Since the enrollment gate is unused, it's removed entirely rather than kept as a flag — former enrollment sites are now `@requires_auth()` (any valid JWT).
- Removed the backend `AUTH0_REQUIRED_PERMISSION` (decorator, `docker-compose.yaml`, `.env.example`) and the frontend `NEXT_PUBLIC_AUTH0_REQUIRED_PERMISSION` gate (`auth0Config.requiredPermission`, the `hasRequiredPermission` context field).
- `requires_auth(required_permission=...)` (e.g. the metrics ADMIN routes) and `check_permissions` are unchanged.

Behavior-preserving: both env vars were already unset in the deployment.

## Testing

- App builds; no `/api/admin*` routes remain (blueprints: `root_api`, `user_api`, `runs_api`, `metrics_api`).
- `ruff` introduced no new findings; UI `ts-check` + eslint clean.
- Verified the metrics ADMIN route still returns 403 without the permission and 200 with it.


dirkraft:
@radamson actually was this jobs_api used only by the admin.html? if so, i should remove it altogether. claude seems to think that is the case

api/admin/jobs_api.py
```diff
@@ -54,17 +51,17 @@ def list_jobs(userid: str):
             return jsonify({"error": str(e)}), 500
 
     @api.route("/create", methods=["POST"])
+    @requires_auth()
     def create_job():
-        """Create a new job."""
+        """Create a new job for the authenticated user."""
+        userid = request.user.get("sub")
```
radamson: Now that `admin.html` has been deleted these routes appear fully unused. It seems like maybe these routes, and maybe entire file, should be deleted unless there is another expected use.

radamson (changes requested):

api/admin/jobs_api.py
dirkraft: yeah somehow i saw "jobs_api" and assumed this was used by the webapp. deleted entirely now

api/user/user_api.py
```diff
@@ -93,7 +93,7 @@ def get_viewer_credits():
 
     # Example protected endpoint - requires special permission
     @api.route("/me/enrollment-status")
-    @requires_enrollment
+    @requires_auth()
     def enrollment_status():  # pyright: ignore reportUnusedFunction
```
radamson, line 97: Probably best in a follow up PR since it would require changes to the debug endpoint, which isn't touched by this PR, but it seems like we can probably remove this enrollment stuff (since all the checks seem to be just for auth now)

radamson (approved):
Looks good, thanks for the cleanup.

Left one comment about enrollment_status, but might be best for a follow-on PR since mostly relates a different set of files (i.e., the user auth debug page rather than the admin page).

api/user/user_api.py
dirkraft: aha good find. yes i'll check this in another pr

PR #42, opened 2026-07-10
add rowIndex for heap analytics

goblinbrooke:
For: https://github.com/allenai/nora-issues/issues/2650
Clicking a hypothesis row in autods now sends a `run_details__experiment-row` event to Heap with a rowindex property (plus experimentId), so we can analyze which rows users click.

I'm also removing the `data-track-name` and updating the tests, because after searching in Heap, I concluded: 
Searching for `run_details__download_btn` (the `data-track-name` event) returned nothing, while the slack thread mentioned `run_details__sort_column` (the `heap.track()` event) shows up. Therefore I removed it and updated the tests so we wouldn't have two mechanisms with only one working.

**Heap testing after merged**
Once deployed and a few rows have been clicked:
1. `run_details__experiment-row` appears in Heap with `rowindex` listed via Custom API (won't show until the first post-deploy click).
2. `rowindex` matches the clicked row's displayed position (0-based); experimentId matches the hypothesis.
3. Confirm sort-aware positioning is the intended semantic (it reflects current sort order, not a fixed experiment number).

### DIFF
--- a/ui/e2e/public-sample.spec.ts
+++ b/ui/e2e/public-sample.spec.ts
@@ -11,7 +11,7 @@ test.describe('Public shared sample', () => {
         await page.goto('/shared/samples/nls_bmi');
 
         // Wait for experiment rows to appear - confirms page fully rendered and API data loaded
-        await page.waitForSelector('[data-track-name="run_details__experiment-row"]', {
+        await page.waitForSelector('.MuiDataGrid-row', {
             timeout: 20000,
         });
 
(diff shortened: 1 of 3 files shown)


AmberRose2 (approved):
FWIW I think my changes adding the wrapper for heap would fix the issue of `run_details__experiment-row` not showing up. - but those heap events won't show up until that code is pushed to prod.
But if this works and is giving more and better data, then I don't think there's a problem with doing it this way!

PR #43, opened 2026-07-10
fix table clipping issue

goblinbrooke:
For: https://github.com/allenai/nora-issues/issues/2931

Matches how the other table components are.

Before: 
Image

After:
Screenshot 2026-07-10 at 1 28 28 PM


AmberRose2:
I think the clipping this ticket is referring to is actually the "Exploration" column causing the text in the column before it to clip?

goblinbrooke:
Clipping is gone now with the exploration tab is there: 
Screenshot 2026-07-10 at 3 40 40 PM

AmberRose2 (approved):

PR #44, opened 2026-07-11
create footer for AD 

goblinbrooke:
For: https://github.com/allenai/nora-issues/issues/3012

Adds the same footer to AD that Asta has. Tested all the footer links.

Before: 
Image

After: 
Desktop: 
Screenshot 2026-07-10 at 5 35 38 PM
Screenshot 2026-07-10 at 5 35 52 PM

Mobile:
Screenshot 2026-07-10 at 5 36 11 PM
Screenshot 2026-07-10 at 5 36 20 PM

### DIFF
--- a/ui/next.config.js
+++ b/ui/next.config.js
@@ -1,7 +1,4 @@
 /** @type {import('next').NextConfig} */
-const nextConfig = {
-    // Hide the Next.js dev-mode indicator so `next dev` mirrors production.
-    devIndicators: false,
-};
+const nextConfig = {};
 
 module.exports = nextConfig;
--- a/ui/next.config.js
+++ b/ui/next.config.js
@@ -1,4 +1,7 @@
 /** @type {import('next').NextConfig} */
-const nextConfig = {}
+const nextConfig = {
+    // Hide the Next.js dev-mode indicator so `next dev` mirrors production.
+    devIndicators: false,
+};
 
-module.exports = nextConfig
+module.exports = nextConfig;
--- a/ui/src/app/components/MenuLinks.tsx
+++ b/ui/src/app/components/MenuLinks.tsx
@@ -39,7 +39,6 @@ export const MenuLinks = () => {
                     href={FEEDBACK_URL}
                     target="_blank"
                     rel="noopener noreferrer"
-                    aria-label="Leave Feedback"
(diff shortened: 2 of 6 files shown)


ui/src/app/components/MenuLinks.tsx
```diff
@@ -0,0 +1,137 @@
...
+                    aria-label="Privacy Policy"
+                    data-test-id={TEST_ID_PRIVACY_POLICY_LINK}
+                    {...mkPrivacyLinkTrackAttrs()}>
+                    Privacy Policy
+                </SettingsLink>
+                <SettingsLink
+                    href="https://allenai.org/terms"
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    aria-label="Terms of Use"
```
ashassan: Doesn't need this cause of the link text. Same with the next one.

ui/src/app/components/MenuLinks.tsx
```diff
@@ -0,0 +1,137 @@
...
+                    type="button"
+                    aria-label="Open disclaimer dialog"
+                    onClick={() => setIsDisclaimerOpen(true)}
+                    data-test-id={TEST_ID_DISCLAIMER_BUTTON}
+                    {...mkDisclaimerBtnTrackAttrs()}>
+                    Disclaimer
+                </SettingsButton>
+                <SettingsButton
+                    type="button"
+                    aria-label="Open attribution dialog"
```
ashassan: you might be able to get rid of this and just use the link text + add  aria-haspopup="dialog" instead

ui/next.config.js
```diff
@@ -1,4 +1,7 @@
 /** @type {import('next').NextConfig} */
-const nextConfig = {}
+const nextConfig = {
+    // Hide the Next.js dev-mode indicator so `next dev` mirrors production.
+    devIndicators: false,
```
ashassan: Should this be done separately? What brought this on?

ashassan (approved):

ui/next.config.js
goblinbrooke: This is to get rid of the Next.JS dev indicator which is on top of the footer button. We spoke offline about it and decided to leave it, so this is being taken out. 
Dev will look like this: 
Image

PR #45, opened 2026-07-16
explore button styling

AmberRose2:
Work for https://github.com/allenai/nora-issues/issues/3022

Changes:

- Put arrow back
- Link is always green and doesn't change color
- Underline on hover
Screenshot 2026-07-16 at 12 34 18 PM


goblinbrooke (approved):
Looks like a link, we love to see it

PR #46, opened 2026-07-17
make intent field the same height as the dataset schema

goblinbrooke:
For: https://github.com/allenai/nora-issues/issues/3020#event-28135379199
The intent field now has 3 lines, same as the dataset schema field

Before: 
Image 

After: 
Screenshot 2026-07-17 at 1 49 25 PM


AmberRose2 (approved):
TY!!!

PR #48, opened 2026-07-20
CI: run @public e2e on PRs and merges to main

dirkraft:
## What

Adds `.github/workflows/e2e-public.yml` — a public-safe e2e job that runs the credential-free **`@public`** Playwright subset against a locally-composed, **bucket-less** stack. Runs on `pull_request` → `main` and `push` → `main`.

A matrix leg per auth provider:
- **`none`** — asserts no sign-in affordance, already authenticated.
- **`password_file`** — login dialog opens; wrong password shows an inline error and does **not** trigger the global "Access Denied" dialog; a seeded user signs in and out.

## Why it's safe for the public repo

Uses **no secrets** (empty `.env`, no GCS bucket or credentials), so it's safe on PRs — including forks. The credentialed integration suite (real Auth0 + GCS + Modal) continues to run on internal infra, triggered from the private repo. This is the "public tier" of the two-tier split we designed.

## How each leg works

1. Empty `.env` (the checked-in symlink is dangling in CI) + bucket-less env (`GCS_BUCKET=""`, `GOOGLE_APPLICATION_CREDENTIALS=/dev/null`).
2. `password_file` only: seed an ephemeral test user via `api/scripts/auth_admin.py`.
3. `docker compose up -d --build proxy`.
4. `yarn playwright install --with-deps chromium` → `yarn test:e2e:public`.

## Validation

Ran both legs locally against a bucket-less stack before writing this:
- Stack builds and boots without GCS; `/api/auth/config` and `/api/user/me` work; `/api/runs/local/list` 500s (expected — no bucket).
- `none` spec: **passed** — `/runs` degrades gracefully despite the runs-list 500 (no white-screen).
- `password_file` specs: **all passed** — dialog / wrong-password / sign-in-out.
- Also confirmed the app boots with an **empty `.env`** (the exact CI condition).

## Notes

- Data-dependent specs (`public-sample`, `runs-pages`, `details-pages`) are intentionally **not** `@public` — they need GCS-backed sample data and stay internal-only.
- First run builds images (~a few min/leg); layer caching can be added later if it matters.

### DIFF
--- a/.gitguardian.yaml
+++ b/.gitguardian.yaml
@@ -7,8 +7,6 @@ version: 2
 
 secret:
   ignored-matches:
-    - name: Public Auth0 SPA client id (NEXT_PUBLIC_*, safe to ship in the browser)
-      match: 6EQ7FtDfVFMdGCWa8SMnGGX3W7p6XVNa
     - name: e2e password_file session secret (throwaway CI value in e2e-public.yml)
       match: ci-e2e-session-secret-not-a-real-secret-0123456789
     - name: e2e test-user password (throwaway CI value in e2e-public.yml)
--- a/.gitguardian.yaml
+++ b/.gitguardian.yaml
@@ -1,0 +1,15 @@
+# GitGuardian / ggshield configuration.
+# https://docs.gitguardian.com/ggshield-docs/reference/configuration
+#
+# Allowlist of known false positives — values that look like secrets but are
+# intentionally public or are throwaway CI test data.
+version: 2
+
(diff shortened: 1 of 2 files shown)


dirkraft:
human comment: this will expand to include other standing e2e tests (there are some 40+ of them) as various cloud-dependencies are replaced with local versions, particularly these two
* gcs buckets -> local host mounted directory
* gcp cloud run jobs -> docker containers or subprocesses

PR #50, opened 2026-07-22
display name footer match asta

goblinbrooke:
For: https://github.com/allenai/nora-issues/issues/3012

Footer display name now matches asta.


ashassan (approved):

goblinbrooke:
While testing this I noticed that AD was getting a console error: 
`React does not recognize the $isBookmarked prop on a DOM element`

So I just added the fix to this PR.

ashassan (approved):

PR #52, opened 2026-07-23
Fix spurious 403 from userid="null" before auth bridge is wired (#51)

### LINKED ISSUE #51: Spurious 403 on run views: UI fires requests with userid="null" before auth bridge is wired (AUTH_PROVIDER=none)
## Summary

With `AUTH_PROVIDER=none`, opening a run (e.g. hard navigation to `/runs/<id>`) briefly shows
**"Request failed with status 403: User cannot view other user's data"**, even for the current
user's own runs. The run actually loads a moment later — the banner is a transient, spurious error.
It does not reproduce under `AUTH_PROVIDER=auth0`.

## What actually happens

The UI issues run/list requests with the literal string `null` as the user id before the current
user id is known:

```
GET /api/runs/null/list             -> 403
GET /api/runs/null/<runid>          -> 403     <- banner comes from here
GET /api/runs/local/<runid>         -> 200     <- succeeds once auth hydrates
```

The server is behaving correctly (`null` is not the owner, not a public user). The bug is
client-side.

## Root cause

`authBridge.getUserId()` defaults to `async () => null` (`ui/src/app/auth/authBridge.ts`); the
active auth provider overwrites it when it mounts. `RunsApi` builds request URLs as
`effectiveUserid = userid ?? await getUserId()` then `encodeURIComponent(effectiveUserid!)`
(`ui/src/app/api/RunsApi.ts`, e.g. the `list`/`getRun` methods). `encodeURIComponent(null)`
produces the string `"null"`, and the `!` non-null assertion hides that it was null.

The window exists only for the `none` provider because of how readiness is reported:

- **`none`** (`ui/src/app/contexts/Auth0Context.tsx`, `NoneAuthProvider`): hardcodes
  `isAuthenticated: true, isLoading: false` synchronously on mount, but wires the `getUserId`
  bridge in a `useEffect` that runs *after* the first commit. So for one render, consumers see
  "auth ready" and fire fetches while `authBridge.getUserId` is still the default `null`.
- **`auth0`** (`Auth0AuthProvider`): `isLoading` stays `true` until an async `init()` resolves the
  session, and `getUserId` reads from the authenticated client. Readiness is coupled to the user id
  being known, so no `null` window.

So `none` decouples its readiness flag from wiring its own `getUserId`; `auth0` couples them.

## Reproduce
1. Run the stack with `AUTH_PROVIDER=none`.
2. Hard-load / directly navigate to `/runs/<an existing run id>`.
3. Observe the 403 banner; note it clears (or the data is present) after auth hydrates. Server logs
   show `/api/runs/null/...` 403 followed by `/api/runs/<user>/...` 200.

## Proposed fix (two layers)
1. **Providers**: wire the `getUserId` bridge synchronously (or keep `isLoading: true` until it is
   wired) in `NoneAuthProvider` (and check `PasswordFileAuthProvider` for the same shape), so
   "ready" never precedes a known user id.
2. **RunsApi (belt-and-suspenders)**: when the resolved user id is null/undefined, short-circuit or
   throw a clear error instead of interpolating `"null"` into the URL. Also clear the transient
   error once a subsequent load succeeds so the banner doesn't latch.

## Acceptance criteria
- Hard-loading a run URL under `AUTH_PROVIDER=none` never issues `/api/runs/null/...` and shows no
  spurious 403.
- No user-id-bearing runs request is sent before the user id is resolved.
- `auth0` behavior unchanged.

## Notes
Pre-existing on `main`, introduced with the swappable auth providers work. UI-only; unrelated to
the job-execution backend.

dirkraft:
Fixes #51.

## Problem

With `AUTH_PROVIDER=none`, hard-navigating to `/runs/<id>` briefly showed **"Request failed with status 403: User cannot view other user's data"** — even for the user's own runs. The UI fired requests with the literal string `null` as the user id before the id was known:

```
GET /api/runs/null/list      -> 403
GET /api/runs/null/<runid>   -> 403   <- banner
GET /api/runs/local/<runid>  -> 200   <- once auth hydrates
```

`encodeURIComponent(null) === "null"`, and a `!` non-null assertion hid that the id was still null. The window existed only for `none` because it reported `isAuthenticated: true` / `isLoading: false` synchronously while wiring its `getUserId` bridge in a `useEffect` that runs after the first commit.

## Fix (two layers)

**1. Wire the `getUserId` bridge before readiness is reported** (`Auth0Context.tsx`)
- `NoneAuthProvider`: wire during render (lazy `useState` initializer) instead of in a post-commit `useEffect`.
- `PasswordFileAuthProvider`: had the same latent race — its bridge effect updated `getUserId` *after* the commit that flips `isAuthenticated` true, so child fetches ran first. Now wired during render.
- `Auth0AuthProvider`: unchanged — already keeps `isLoading: true` until the user id is known.

**2. Guard `RunsApi` (belt-and-suspenders)** (`BaseApi.ts`, `RunsApi.ts`)
- `BaseApi.requireUserId()` throws a clear error when the id is null/empty instead of interpolating `"null"` into the URL. Removed all `effectiveUserid!` / `userid!` assertions (11 call sites).

## Verification
- Added a `@public` regression e2e test (`auth-providers.spec.ts`, `none` block) using a synthetic, data-independent run id. **Fails without the fix** (captures `/api/runs/null/...`), **passes with it**.
- Reproduced and verified against the live `none` stack; full `none` suite passes; `yarn ts-check` and `yarn lint` clean.

Acceptance criteria met: no `/api/runs/null/...` under `AUTH_PROVIDER=none`, no user-id-bearing request before the id resolves, `auth0` behavior unchanged.

### DIFF
--- a/ui/e2e/auth-providers.spec.ts
+++ b/ui/e2e/auth-providers.spec.ts
@@ -88,4 +88,41 @@ test.describe('Auth: none provider (desktop mode)', { tag: '@public' }, () => {
         // And the logged-out intro CTA is not shown.
         await expect(page.getByRole('button', { name: /Sign in to get started/i })).toHaveCount(0);
     });
+
+    // Regression for issue #51: on a hard navigation to a run, the UI fired
+    // user-scoped requests with the literal string "null" as the user id before
+    // the auth bridge was wired (encodeURIComponent(null) === "null"), producing
+    // a spurious 403. Uses a synthetic run id so the test is data-independent —
+    // the bug fired the null request regardless of whether the run exists.
+    test('hard navigation to a run never issues /api/runs/null/...', async ({ page }) => {
+        const nullUserRequests: string[] = [];
+        const userIdSegments: string[] = [];
+        page.on('request', (req) => {
+            const { pathname } = new URL(req.url());
+            if (/^\/api\/runs\/null(\/|$)/.test(pathname)) {
+                nullUserRequests.push(pathname);
+            }
+            const match = pathname.match(/^\/api\/runs\/([^/]+)\//);
+            if (match && match[1] !== 'shared') {
+                userIdSegments.push(match[1]);
+            }
+        });
+
+        await page.goto('/runs/00000000-0000-0000-0000-000000000000', {
+            waitUntil: 'domcontentloaded',
+        });
+        // Wait until at least one user-scoped runs request has resolved, then a
+        // short buffer, so we observe what the page fired once auth hydrated.
+        await page.waitForResponse((r) => /^\/api\/runs\//.test(new URL(r.url()).pathname), {
+            timeout: 15000,
+        });
+        await page.waitForTimeout(500);
+
+        expect(
+            nullUserRequests,
+            `spurious "null" user-id requests: ${nullUserRequests.join(', ')}`
+        ).toEqual([]);
+        // The resolved id was the desktop local user, proving auth was ready.
+        expect(userIdSegments).toContain('local');
+    });
 });
--- a/ui/src/app/api/BaseApi.ts
+++ b/ui/src/app/api/BaseApi.ts
@@ -18,6 +18,21 @@ export class BaseApi {
         }
     };
 
+    /**
+     * Resolve the user id for a user-scoped request, preferring an explicitly
+     * provided id and falling back to the active auth provider. Throws when the
+     * id is not yet known so callers never interpolate the literal string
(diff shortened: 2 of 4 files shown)


dirkraft:
Manual testing:
* load run details with AUTH_PROVIDER=none
* with AUTH_PROVIDER=password_file
* not logged in, looking at sample run
* not logged in, look at private run (see login button)

PR #53, opened 2026-07-23
Scope Docker job GCS mount to the run's own prefix (#49)

### LINKED ISSUE #49: Scope the job GCS mount to the run's own prefix instead of the whole bucket
## Summary

AutoDiscovery job processes mount the **entire** GCS bucket at `/mnt/gcs`, but a job only ever
reads/writes its own `users/<uid>/jobs/<jid>` prefix. Scope the mount to that prefix (or, as a
looser option, the user's area `users/<uid>`) so a job container cannot see or modify other users'
data through the filesystem. This should apply to **both** job backends — local Docker containers
and Cloud Run executions.

## Current behavior

- **Docker backend**: the job container self-mounts the whole bucket via
  `gcsfuse --implicit-dirs $GCSFUSE_BUCKET /mnt/gcs` (`packages/autodiscovery/docker-entrypoint.sh`).
- **Cloud Run backend**: the job's Cloud Run definition mounts the whole bucket
  (`--add-volume type=cloud-storage,bucket=... --add-volume-mount mount-path=/mnt/gcs`, see
  `packages/autodiscovery/scripts/rebuild_and_deploy.sh`).

The job binary has **no** bucket-wide paths — it only touches what it is handed via args
(`--dataset_metadata`, `--out_dir`, `--work_dir`, `--bucket_path`), all under
`users/<uid>/jobs/<jid>`. So the whole-bucket mount is inherited from the Cloud Run volume model
(and copied to the Docker backend for arg parity), not a functional requirement.

## Motivation

Least privilege / tenant isolation. Today anything in the job container can read and write every
user's data via `/mnt/gcs`. Narrowing the mount removes cross-tenant filesystem access. This
matters most for multi-user / self-hosted deployments. (Note: the Modal code-execution sandbox is
already scoped to the per-job `.../data` prefix, read-only — this issue is about the *job
container's* `/mnt/gcs` mount, which is separate.)

### Docker backend (straightforward)
Mount only the run's prefix, at the matching deep mountpoint so existing job args still resolve:

```sh
mkdir -p "/mnt/gcs/${ONLY_DIR}"
gcsfuse --implicit-dirs --only-dir "${ONLY_DIR}" "${GCSFUSE_BUCKET}" "/mnt/gcs/${ONLY_DIR}"
# ONLY_DIR = users/<uid>/jobs/<jid>   (or users/<uid> for the user-area option)
```

`DockerBackend` passes the prefix via a new env var; the entrypoint mounts the scoped subtree at
`/mnt/gcs/users/<uid>/jobs/<jid>`. Because the subtree is mounted at its full path, the job's
existing absolute paths still resolve and `build_job_args` stays unchanged.

### Cloud Run backend (needs a design decision)
Cloud Storage FUSE volumes support an `only-dir` mount option, but the volume is defined on the
**shared** Cloud Run job resource, and `RunJobRequest` overrides only cover container args/env —
not volumes. So a single job definition can't carry a per-run `only-dir`. Options to evaluate:

1. Create/patch a per-run job definition with the run's `only-dir` (adds latency + lifecycle
   management).
2. Scope to a coarser, static level that the shared definition can carry (does not achieve per-run
   isolation on its own).
3. Move the job off the FUSE mount for its own prefix and access GCS via the SDK scoped to the
   run's prefix (larger change).

The right approach here is the main open question for this ticket.

## Notes / gotchas
- Prefer **job-directory** scope; **user-area** (`users/<uid>`) is an acceptable middle ground.
- User ids can contain a `|` character (a valid path/prefix char) — quote carefully in the
  entrypoint and the `only-dir` value.
- Keep `build_job_args` identical across backends where possible.

## Acceptance criteria
- A job container can read/write only its own prefix (or user area) via `/mnt/gcs`; access to other
  prefixes is not possible from the container.
- Behavior verified for the Docker backend.
- Cloud Run backend either implements equivalent scoping or documents the chosen approach/limitation.

## Context
Split out of the swappable job backend work (#30) as its own change, since it applies to both
backends and is independent of the launcher abstraction.

dirkraft:
Closes #49.

## Why
The Docker job backend self-mounts the GCS bucket at `/mnt/gcs` via gcsfuse, mounting the **whole** bucket. A job only ever reads/writes its own `users/<uid>/jobs/<jid>` prefix, so the whole-bucket mount lets any job container read/write every user's data through the filesystem. This scopes the mount to the run's own prefix (least privilege / tenant isolation).

> Stacked on `swappable-job-backend` (#30), since that's where the Docker backend and entrypoint live. Merge after / retarget once #30 lands.

## Docker backend (scoped)
- `DockerBackend` passes `GCSFUSE_ONLY_DIR=users/<uid>/jobs/<jid>`.
- `docker-entrypoint.sh` mounts the scoped subtree at its **full path** (`/mnt/gcs/$GCSFUSE_ONLY_DIR`) with `gcsfuse --implicit-dirs --only-dir`, so the job's existing absolute `--dataset_metadata` / `--out_dir` args still resolve and `build_job_args` is unchanged.
- `GCSFUSE_ONLY_DIR` is quoted everywhere in the entrypoint — user ids can contain `|`, a valid path char that's shell-special unquoted. Added a test with a `|`-containing user id.

## Cloud Run backend (documented limitation)
Per the issue's open design question: the FUSE volume is defined on the **shared** Cloud Run job resource, and per-execution `RunJobRequest.Overrides` cover only `container_overrides` (args/env), `task_count`, and `timeout` — **not** volumes or a `only-dir` mount option (verified against `google-cloud-run`). So a single shared job definition can't carry a per-run `only-dir`; real per-run scoping would require a distinct per-run job definition (create/patch — added latency + lifecycle). This is documented inline in `gcp.py` and `scripts/rebuild_and_deploy.sh` rather than implemented. (The Modal code-execution sandbox is separately scoped read-only to the per-job `.../data` prefix — out of scope here.)

## Acceptance criteria
- [x] Docker: job container can read/write only its own prefix via `/mnt/gcs`; other prefixes are not visible.
- [x] Cloud Run: chosen approach/limitation documented.

## Testing
- `pytest tests/test_backends.py` — 21 passed (adds scoped-mount assertion + `|`-userid case).
- `ruff check` / `ruff format --check` clean; `sh -n` / `bash -n` on both scripts pass.

### DIFF
--- a/packages/autodiscovery/docker-entrypoint.sh
+++ b/packages/autodiscovery/docker-entrypoint.sh
@@ -9,12 +9,26 @@
 # gs://$GCSFUSE_BUCKET at /mnt/gcs (authenticating via GOOGLE_APPLICATION_CREDENTIALS)
 # before the job starts. gcsfuse requires the /dev/fuse device and CAP_SYS_ADMIN,
 # which the Docker backend grants when launching the container.
+#
+# When GCSFUSE_ONLY_DIR is also set, the mount is scoped to just that prefix
+# (this run's users/<uid>/jobs/<jid>) for least-privilege tenant isolation: the
+# container then cannot see or modify other users' data through the filesystem.
+# The scoped subtree is mounted at its full path (/mnt/gcs/$GCSFUSE_ONLY_DIR) so
+# the job's existing absolute --dataset_metadata / --out_dir args still resolve.
 set -e
 
 if [ -n "${GCSFUSE_BUCKET}" ]; then
-    echo "Mounting gs://${GCSFUSE_BUCKET} at /mnt/gcs via gcsfuse..."
-    mkdir -p /mnt/gcs
-    gcsfuse --implicit-dirs "${GCSFUSE_BUCKET}" /mnt/gcs
+    if [ -n "${GCSFUSE_ONLY_DIR}" ]; then
+        # Quote GCSFUSE_ONLY_DIR everywhere: user ids can contain a '|' char,
+        # which is a valid path character but shell-special when unquoted.
+        echo "Mounting gs://${GCSFUSE_BUCKET}/${GCSFUSE_ONLY_DIR} at /mnt/gcs/${GCSFUSE_ONLY_DIR} via gcsfuse..."
+        mkdir -p "/mnt/gcs/${GCSFUSE_ONLY_DIR}"
+        gcsfuse --implicit-dirs --only-dir "${GCSFUSE_ONLY_DIR}" "${GCSFUSE_BUCKET}" "/mnt/gcs/${GCSFUSE_ONLY_DIR}"
+    else
+        echo "Mounting gs://${GCSFUSE_BUCKET} at /mnt/gcs via gcsfuse..."
+        mkdir -p /mnt/gcs
+        gcsfuse --implicit-dirs "${GCSFUSE_BUCKET}" /mnt/gcs
+    fi
 fi
 
 exec /root/.local/bin/uv run python -m autodiscovery.run "$@"
--- a/packages/autodiscovery/scripts/rebuild_and_deploy.sh
+++ b/packages/autodiscovery/scripts/rebuild_and_deploy.sh
@@ -88,6 +88,13 @@ UPDATE_SECRETS=(
 )
 
 UPDATE_SECRETS_ARG=$(IFS=,; echo "${UPDATE_SECRETS[*]}")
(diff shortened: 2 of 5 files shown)


dirkraft:
GCP cloud run jobs don't allow per invocation-specific volume mounts. So this would only work for docker containers. I don't like the asymmetry here, and the amount of comments in the code beget that asymmetry. Not worth the cognitive load.

Issue #55, opened 2026-07-23
Unify AD-job code execution onto the asta-sandbox backend registry


dirkraft:
### Context

Making the code-execution backend configurable (part of #34) surfaced an inconsistency in how the AD job runs generated code. The job's `--backend` flag has three values, but only one routes through `asta-sandbox`:

- `modal` → `asta_sandbox.backends.modal_ephemeral.ModalEphemeralExecutor` (asta-sandbox)
- `process` → the workspace's own `code_execution.ProcessIPythonBackend` (**not** asta-sandbox), wrapped by an ad-hoc `_ProcessBackendAdapter` in `packages/autodiscovery/src/autodiscovery/agents.py`
- `local` → autogen's `LocalCommandLineCodeExecutor` (**not** asta-sandbox)

`asta-sandbox` already ships equivalent backends: `inprocess`, `docker`, `modal_ephemeral`, `modal_kernel`. So today's `process`/`local` are parallel, one-off implementations of things the library provides (or could).

### Proposal

Route all code execution through the `asta-sandbox` backend registry so backend selection is uniform and the ad-hoc adapter goes away:

- Transplant `ProcessIPythonBackend` onto `asta_sandbox.SandboxBase` as a first-class local process backend (make `run_cell` async via `asyncio.to_thread`) and move it into the `asta-sandbox` library alongside `InProcessExecutor` and `ModalEphemeralExecutor`. This is already flagged by a `TODO` in `packages/code_execution/src/code_execution/process_backend.py`.
- Once it lives in the library, delete `_ProcessBackendAdapter` in `agents.py` and construct executors from a single `SandboxBase` factory keyed by the backend name.
- Consider mapping the job's `--backend` names onto the library's names (or adopting the library's names directly) so the vocabulary is shared.

### Why not now

This PR keeps the change small and low-risk: it makes the existing, working knob configurable (`CODE_EXECUTION_BACKEND`) without moving code between packages or changing the `asta-sandbox` public surface. The unification is a mechanical-but-cross-package refactor best done on its own.

### Acceptance

- The AD job's code execution is selected via a single `asta-sandbox` `SandboxBase` factory.
- `_ProcessBackendAdapter` is removed.
- No behavior change for existing `process`/`local`/`modal` runs.

Related: #34

gas2own:
[...]

Issue #54, opened 2026-07-23
improved configuration syntax


dirkraft:
It's all env vars and its hard to see which become required depending on which auth provider is active, or job backend, or other dependent settings.

dirkraft:
### Cross-field validation opportunity: code-execution backend × job backend × tenancy

While making the code-execution sandbox configurable, we hit a case that today can only be documented and (loosely) warned about, but that a richer config model could *validate* and reject outright.

The interacting fields:

- **Job backend** (`JOB_BACKEND`): `docker` | `gcp`
- **Code-execution backend** (`CODE_EXECUTION_BACKEND`): `process`/`local` (in the job container) | `modal` (scoped, read-only, isolated)
- **Auth/tenancy** (`AUTH_PROVIDER`): `none` (single local user) vs `auth0`/`password_file` (multi-user)

The hazard: the AD job container mounts GCS data for the run, and `process`/`local` execute untrusted, LLM-generated code **inside that container**, so the code sees whatever the container mounts.

- `docker` backend can gcsfuse-mount *only* the job's own prefix per-invocation → in-process is safely scoped even multi-tenant.
- `gcp`/Cloud Run cannot vary the mount per execution (one job resource, fixed mount) → the whole bucket stays mounted → in-process code in a multi-tenant deployment can read **all users' data**.

So the combination `JOB_BACKEND=gcp` + `CODE_EXECUTION_BACKEND` in {`process`,`local`} + `AUTH_PROVIDER != none` is a cross-tenant data-exposure footgun. For now we're handling it with a **loud startup warning** + docs (a safe-combination matrix); env vars can't express "these three fields are only valid in certain combinations."

When we take on improved configuration syntax, this is a concrete motivating example for **cross-field validation** — the config model should be able to declare invalid combinations and refuse them at load time (turning the current soft warning into a hard, self-documenting constraint). Related: #34.

dirkraft:
In this work we should also be consistent about how configuration is referenced throughout the codebase. i.e. scattered os.environ.get and equivalent calls

Issue #34, opened 2026-07-07
Decouple from Ai2-specific vendors


dirkraft:
This project powers asta-autodiscovery.allen.ai where a number of specific cloud providers have been selected for Ai2's needs. To make this codebase more portable and adaptable to arbitrary environments, these cloud-specific implementations need to be abstracted and replaced with low/no-barrier default options.

* [x] Authentication - Abstract to three stock implementations: Auth0, password file, no auth (local mode, new default)
  * https://github.com/allenai/asta-autodiscovery/pull/29
* [x] AutoDiscovery jobs - Abstract to two stock implementations: GCP Cloud Run, docker containers (new default)
  * https://github.com/allenai/asta-autodiscovery/pull/30
* [x] Code execution sandboxing - Enable configuration of backends supported by https://pypi.org/project/asta-sandbox/ on which this project depends.
  - https://github.com/allenai/asta-autodiscovery/pull/56
* [x] LLM providers #68 
* [ ] Persistence - Abstract to two stock implementations: GCS, docker host mount (new default)
* [ ] Compute credit tracking - Option to disable, which will become the default. Maybe abstract an interface so 3rd-party operators can implement their own bookkeeping. 
  - 7/23 plan: organize asta integration code into asta subpackages. Option to disable UI elements. Backend balance checks either disabled, or satisfied by "infinite" credits, whichever solutions is more elegant, fewer unique code paths/LoC.
* [ ] Enable public repo integration tests - a lot of tests don't run because they depend on private infra but after this point all(?) will be runnable using various local, non-cloud components
* [ ] #54 
* [ ] AutoDiscovery -> Asta DataVoyager integration - Agent-to-agent handoff is critical to research workflows, but how can we rationalize this from a systems point of view when Ai2 isn't the operator of both agent systems in that exchange?
* [ ] isolate asta-specific integration points in asta-specific subpackages
* ... probably others

PR #56, opened 2026-07-23
Configurable code-execution backend (CODE_EXECUTION_BACKEND)

dirkraft:
Part of #34. Makes the AutoDiscovery job's code-execution sandbox configurable, so the low-barrier default no longer depends on a specific cloud vendor.

## What

The AD job already accepts `--backend {local,process,modal}`; the webstack just hardcoded `modal`. This surfaces that knob as `CODE_EXECUTION_BACKEND`, mirroring the `JOB_BACKEND` pattern, and defaults to **`process`** — an isolated subprocess inside the job container with no cloud dependency.

- `JobConfig.code_execution_backend` (env `CODE_EXECUTION_BACKEND`).
- `build_job_args` emits `--backend=<value>`; `--bucket_path` only for `modal` (it's the modal data mount).
- Docker backend **scopes the per-job gcsfuse mount** to `users/<uid>/jobs/<jid>` (`GCSFUSE_ONLY_DIR`), mounted at the same deep path so the args stay backend-agnostic.
- Entrypoint honors `GCSFUSE_ONLY_DIR` (whole-bucket fallback otherwise).
- Loud startup warning for the one unsafe combination (below).
- Docs: safe-combination matrix; `MODAL_*` now required only when `CODE_EXECUTION_BACKEND=modal`.

## The security angle (data scoping)

In-process backends (`process`/`local`) run untrusted, generated code **inside the job container**, so it sees whatever GCS data the container mounts. The sandbox's real value is *scoping* the data view, not just isolating the process.

| Job backend | Code backend | Single user | Multi-user |
| --- | --- | --- | --- |
| `docker` | `process`/`local` | ✅ | ✅ mount scoped to the job's own prefix |
| `gcp` | `process`/`local` | ✅ | ⚠️ **unsafe** — Cloud Run mounts the whole bucket, can't scope per execution |
| `docker`/`gcp` | `modal` | ✅ | ✅ scoped, read-only per-job mount |

The docker backend self-mounts per container, so it *can* scope per run; Cloud Run's mount is fixed per job, so it can't. Multi-user Cloud Run deployments must use `modal`; the API warns loudly if misconfigured.

## Testing

`autodiscovery_jobs` suite green (77 passed), including new coverage for the arg branching, the scoped mount env, and the unsafe-combo warning matrix.

## Follow-ups
- #55 — unify `process`/`local` onto the asta-sandbox backend registry.
- #54 — cross-field config validation could turn the soft warning into a hard constraint.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,13 +22,27 @@ JOB_BACKEND=docker
 # `docker compose build autodiscovery` (tags it as this value).
 AUTODISCOVERY_IMAGE=autodiscovery:dev
 
+# Code-execution backend for the AD job's LLM-generated code:
+#   process (default) - isolated subprocess inside the job container; no cloud
+#                       dependency (Modal not required). Reads the dataset from
+#                       the job's /mnt/gcs mount.
+#   local             - in-process, no isolation.
+#   modal             - remote Modal sandbox; mounts only the per-job data,
+#                       read-only. Required for multi-user GCP deployments.
+# Untrusted generated code runs wherever the data is mounted, so this interacts
+# with JOB_BACKEND and AUTH_PROVIDER. See docs/configuration.md for the safe
+# combinations before changing it for a multi-user deployment.
+CODE_EXECUTION_BACKEND=process
+
 MODAL_TOKEN_ID=...
 MODAL_TOKEN_SECRET=...
 MODAL_IMAGE_BUILDER_VERSION=2025.06
 MODAL_ENVIRONMENT=AutoDiscovery
 # Name of the Modal secret holding the GCS credentials the sandbox uses to mount
-# datasets. Required: must name a real secret in MODAL_ENVIRONMENT, or every run
-# fails at sandbox creation with "Secret '...' not found".
+# datasets. Required only when CODE_EXECUTION_BACKEND=modal: it must name a real
+# secret in MODAL_ENVIRONMENT, or every run fails at sandbox creation with
+# "Secret '...' not found". Unused (and the MODAL_* vars above are unused) with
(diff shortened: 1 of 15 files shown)


dirkraft:
I did a spot check myself with two different configurations

* cut out as much of Ai2's dependencies as one currently can
```
export AUTH_PROVIDER=none
export JOB_BACKEND=docker
export CODE_EXECUTION_BACKEND=process
...
```

* my standing configuration which tests Ai2 infra and includes
```
export AUTH_PROVIDER=auth0
export JOB_BACKEND=gcp
export CODE_EXECUTION_BACKEND=modal
...
```

dirkraft:
@rodneykinney @charliemcgrady tagging you two on this. Feel free to look as closely or not as you'd like. Tagging y'all because it's inthe general vein of thinking about process-level components and the seams between them, and could stoke some of our other larger discussions.

In this case, the seam is between an AD job which generates code, and the actual execution of that code. 
* On autodiscovery.allen.ai, it's gcp cloud run jobs -> modal sandbox. Both get cloud bucket mounts.
* Here in this PR that seam becomes very thin: AD job docker container -> execute code in same container. Acceptable for two reasons
  * an AD job container mounts a more specific bucket path - this isn't possible in gcp **per invocation** so AD jobs actually have visibility of the entire bucket across all users' jobs. When we launch modal sandboxes for code execution they get user-job-specific mountpoints and so there's no risk of hijacking other data.
  * running the AD stack like this is akin to "desktop mode" so we generally accept that applications running on our own machine can access much of the local machine

rodneykinney (approved):

PR #58, opened 2026-07-30
Swappable persistence backend: docker host mount + GCS (#34)

dirkraft:
## Context

Part of #34. Abstracts **persistence** behind an object-store interface with two stock
implementations, and makes the vendor-free one the default — following the same shape as
[#29](https://github.com/allenai/asta-autodiscovery/pull/29) (auth), [#30](https://github.com/allenai/asta-autodiscovery/pull/30) (job backend), and [#56](https://github.com/allenai/asta-autodiscovery/pull/56) (code execution).

- **`local`** (new default) — a directory on the host, bind-mounted into the containers. No
  cloud account, bucket, or credentials; `docker compose up` persists real run data out of
  the box.
- **`gcs`** — the existing Cloud Storage bucket, unchanged behavior when selected.

Both backends use the **same key layout**, so a bucket and a data directory are
interchangeable (`gsutil rsync` moves data either way).

Design writeup: `docs/design/storage-backends.md`.

## Changes

**New `autodiscovery_jobs/storage/` subpackage**
- `base.py` — `ObjectStore` ABC, `ObjectInfo`, `glob_to_regex()`
- `gcs.py` — `GcsStore`
- `local.py` — `LocalStore`
- `__init__.py` — `get_store(config)` factory

**Wiring**
- `gcs.py` (~1100 lines) → `persistence.py` over the store; `gcs.py` becomes a thin compat
  shim re-exporting it, mirroring the `cloudrun.py` precedent from #30
- `run_details.py` / `user_profile.py` / `email_state.py` / `asta_gcs.py` → store
- `api/metrics/aggregator.py` → store (glob discovery + persisted cache snapshot)
- `scripts/send_completion_emails.py` lock → `create_exclusive`
- `api/runs/runs_api.py` — upload URL fallback + preloaded-dataset sync via the store
- `config.py` → `storage_backend` (`STORAGE_BACKEND`) + `storage_dir` (`STORAGE_DIR`)
- docker-compose / `.env.example` / `docs/configuration.md` / proxy body-size comments

**Docker job backend** bind-mounts the run's own host subtree instead of gcsfuse-mounting
it, at the same in-container path (now the constant `JOB_MOUNT_ROOT`), so the job's CLI args
are byte-identical across backends. The bind case grants no `/dev/fuse`, no `SYS_ADMIN`, and
no credentials.

## Making a filesystem behave like an object store

Four GCS behaviors the existing code already depended on:

| GCS behavior | `LocalStore` |
| --- | --- |
| `matchGlob`, where `*` does not cross `/` | `glob_to_regex()` (`*` → `[^/]*`, `**` → `.*`). Deliberately **not** `fnmatch`, whose `*` crosses `/` and would make `users/*/jobs/*/x.json` match arbitrarily deep keys |
| No directories — a prefix stops existing once its last object is deleted | deletes prune the directories they empty; prefix listings skip directories with no objects beneath |
| Writes are atomic; a reader never sees a partial object | staged `.ad-staging.*` file + `os.replace`, and listings skip staging files. Matters because the API polls run files *while the job container writes them* |
| `if_generation_match=0` | `os.open(..., O_CREAT \| O_EXCL)` |

`created_at` comes from `st_mtime` (no portable POSIX creation time); these objects are
written once, so it's the same instant in practice. Keys can come from user-supplied
filenames, so any key resolving outside the root is rejected rather than normalized.

## Callouts for review

**1. GCS deployments must now set `STORAGE_BACKEND=gcs` explicitly.** Same situation as #30
changing `JOB_BACKEND`'s default, but with a sharper failure mode, which is why:

**2. Two combinations fail at startup rather than warning.** `local` + `JOB_BACKEND=gcp`
(Cloud Run can't mount a host directory) and `local` + `CODE_EXECUTION_BACKEND=modal` (the
sandbox mounts the dataset from `gs://`). Neither can work. The likeliest way to hit the
first is a GCS deployment that never set `STORAGE_BACKEND` — silently defaulting to local
disk there would look like every run and dataset had vanished, which is worse than failing
to boot. Contrast `_warn_if_unsafe_code_execution`, which only warns because that config
*works*, it's just unsafe. **Push back if you'd rather this warn.**

**3. Uploads go through the API on the local backend.** A filesystem has no presigned-URL
equivalent, so `generate-upload-url` returns a same-origin
`PUT /api/runs/<runid>/datasets/<filename>` that streams `request.stream` into the store
(no memory buffering, no temp file). The new `same_origin` response flag tells the UI to
attach its bearer token — which it must **not** send to `storage.googleapis.com`, since
that would leak it. This also makes the proxy's `client_max_body_size` for `/api`
load-bearing again (its `REMOVE AFTER PRESIGNED UPLOADS ARE IMPLEMENTED` comment is now
wrong; updated rather than removed).

**4. The in-container mount point stays `/mnt/gcs`, even for the local backend.** The
deployed Cloud Run job definition pins that path (`--add-volume-mount` in
`rebuild_and_deploy.sh`), so renaming it would couple this PR to a Cloud Run redeploy — a
deploy-ordering hazard I didn't want here. It's now the single constant
`backends.base.JOB_MOUNT_ROOT`, so renaming it later is one line plus a redeploy. Flagged as
follow-up in the design doc.

**5. Wire-format compatibility kept where it was cheap.** `GCSError` is now an alias of the
new `StorageError` (existing `except GCSError` sites keep working), and the
`generate-upload-url` response keeps its `gcs_path` field name — now carrying a `file://`
URI under the local backend — rather than breaking the UI contract.

**6. The metrics aggregator is in scope** even though "Compute credit tracking" is a
separate #34 checkbox. It's ~6 functions threading a `storage.Bucket` around, and leaving it
would mean the new default silently can't render metrics.

**7. `asta_gcs.py` still writes to GCS** — that bucket is Asta's, not ours — but now reads
its source through the store, doing a server-side copy only when the source is also GCS.

## Verification

**Ran the real stack with no GCS credentials at all** (`STORAGE_BACKEND` unset → local,
`AUTH_PROVIDER=none`, `JOB_BACKEND=docker`, `CODE_EXECUTION_BACKEND=process`):

- created runs, uploaded a dataset through the new same-origin endpoint, saved metadata,
  listed and fetched runs, refreshed status, soft-deleted — all round-tripping on disk with
  the expected layout
- the metrics aggregator scanned the local store via the glob listing and persisted
  `_metrics/jobs_cache.json`
- submitted a run: the job container launched with exactly one scoped bind mount
  (`data/users/local/jobs/<id>` → `/mnt/gcs/users/local/jobs/<id>`), `CapAdd=[]`,
  `Devices=[]`, and no `GCS*`/`GOOGLE*`/`MODAL*` env — then **read `metadata.json` and wrote
  `output/args.json` + `llm_usage_events.jsonl` back through the mount**

Two unrelated failures surfaced while doing that, neither caused by this change and neither
fixed here:
- the run then died on `VERTEX_PROJECT_ID` being unset in my env, because the default model
  is a Gemini model. Arguably another Ai2-vendor coupling for #34, but a different checkbox.
- an earlier attempt crashed in `autodiscovery/dataset.py:get_asta_description` because my
  metadata omitted `description` (`"\n".join` over a `None`) — a pre-existing
  input-validation gap.

**Automated:** 310 tests pass (store contract, GCS glob parity, path-traversal
rejection, prefix pruning, concurrent same-key writes, and the persistence API end to end
against a real temp dir). `test_gcs.py`'s low-level client mocks were replaced by real
round-trips against `LocalStore`, plus a focused `GcsStore` translation-layer suite.
Repo-wide pyright: **126 errors vs 256 on main**. ruff: 81 findings vs 86 on main over the
same paths. UI `tsc` and `next lint` clean.

## Merged up to main (`c326b94`)

Now current with 40 commits of main. Most conflicts were mechanical, but three of main's
changes landed *inside* code this PR rewrites and had to be carried across the refactor:

- **#80's per-user GDPR purge** was added to `gcs.py`, the file this PR guts into a
  re-export shim. Ported to `persistence.py` over the `ObjectStore` interface — so erasure
  now works on **any** backend instead of GCS only. `UserDataSummary.bucket` becomes
  `.location` (the store root URI) and `purge_user_data` returns `"location"` in place of
  `"bucket"`; `scripts/purge_user_data.py` updated to match. **This is the one intentional
  API change to a thing main shipped — flag it if you'd rather keep the `bucket` name.**
- **#76's `get_metadata_or_none`** (the single-round-trip form credit aggregation uses)
  didn't exist on this branch. Extracted from `persistence.get_metadata` with the same
  contract.
- **#91's** metrics discovery failure now propagates instead of reading as an empty
  dataset. Kept as main wrote it, with the message naming `store.root_uri` rather than
  `config.bucket`.

Two of main's test files were written against GCS wire mocks the store interface no longer
matches. `test_user_purge.py` is rewritten against a real `FilesystemStore` on a temp dir
(this PR's existing pattern) and additionally asserts the purge stays scoped to one
subject; `test_metrics_aggregator.py` is repointed at the `get_store` seam.

Resolved as deletions: `ui/src/app/runs/actions.ts` (main deleted the whole file in #68;
this branch had only pruned its unused upload helper) and `tests/test_gcs.py` (replaced
here by `test_storage.py` / `test_persistence.py`).


dirkraft:
Follow-up pushed (38382e5) from the review question *"what would a third provider have to implement, and is that even possible?"*

The honest answer to the second half was **not really** — you could write the class, but it would have to edit four places, and two of them failed *silently*. Fixed:

**1. Capabilities instead of name comparisons.** `ObjectStore` now declares two things, both defaulting to "no":

```python
job_data_mount: JobDataMount = JobDataMount.UNSUPPORTED  # HOST_PATH | GCSFUSE | UNSUPPORTED
gs_addressable: bool = False
```

`docker.py` and `_validate_storage_config` ask the store class rather than comparing `storage_backend` strings. The old `needs_fuse = storage_backend == "gcs"` sent everything else to the bind-mount branch, so a hypothetical S3 backend would have launched a job against a host path that doesn't exist; it's now refused with a message naming the missing declaration. Capabilities are read from the **class**, not an instance, so validation doesn't `mkdir` a possibly-unwritable `STORAGE_DIR` as a side effect. The factory became a name→class registry to support that lookup.

`gs_addressable` is narrower than "is remote" on purpose — Cloud Run volumes and Modal's `--bucket_path` understand GCS specifically, so an S3 store is remote *and* still unusable by them.

**2. Smaller required surface.** `upload_file`, `download_file`, and `copy` are now defaulted from the primitives: **12 abstract members → 9**. Both real stores still override all three natively (GCS copies server-side, which matters since forking a run copies its whole dataset).

**3. `LocalStore` → `FilesystemStore`.** The name undersold it. `FilesystemStore` is a POSIX-tree backend, so **the zero-code path to other storage is: mount it and use `local`** — NFS, s3fs, Azure Files, JuiceFS. Job containers included; they just get a bind mount of the run's subdirectory. The configured value stays `local`, which is still accurate about the constraint that matters (host-only reachability is exactly what rules out Cloud Run and Modal).

**Docs** now spell out both tiers in [Adding a third backend](https://github.com/allenai/asta-autodiscovery/blob/swappable-persistence-backends/docs/design/storage-backends.md#adding-a-third-backend), including what Tier 1 costs you, since it isn't nothing:

| | Cost on a plain mount |
|---|---|
| Presigned uploads | gone — uploads stream through the API |
| `list(match_glob=…)` | one request against GCS becomes one readdir per directory |
| `copy` | no server-side copy; forking a run round-trips the dataset through the API |
| `create_exclusive` | `O_EXCL` isn't a cross-machine guarantee over a FUSE object-store adapter — the completion-email lock quietly stops being a lock |
| atomic replace | same caveat: rename is typically copy+delete there, so a reader can see a partial object |

For a single-operator deployment most of those don't bite. Registration is still a source edit (no entry points) — deliberate stopping point, noted as such in the doc.

**Tests:** +13. A dict-backed `MinimalStore` implements exactly the nine members with the functional persistence layer run against it, which is the executable statement of the contract; plus the docker backend refusing an `UNSUPPORTED` store, and the class-not-instance capability lookup.

Re-ran the stack end to end afterward: mount is still the single scoped bind (`CapAdd=[]`, `Devices=[]`), and the job still reads `metadata.json` and writes `args.json` + `llm_usage_events.jsonl` through it. 236 tests pass; pyright still at the 3 pre-existing errors.

api/runs/runs_api.py
```diff
@@ -115,33 +117,47 @@
...
+        rewrite_token = None
+        while True:
+            rewrite_token, bytes_rewritten, total_bytes = dest_blob.rewrite(
+                source_blob, token=rewrite_token
+            )
+            if rewrite_token is None:
+                break
+    else:
+        with source_blob.open("rb") as stream:
+            dest_store.write_stream(dest_key, stream)
```
dirkraft, line 156: This is super leaky. I don't want to address now since it assumes GCS->GCS but leave a # TODO somewhere in here, or perhaps we should just throw "not supported" if we're not in a GCS->GCS configuration

dirkraft:
Pushed a586b18 — you were right, and I don't have a defense. I added `PUT /api/runs/<runid>/datasets/<filename>` without checking that `POST /api/runs/upload-dataset` already existed, was already authenticated, and already wrote through the same persistence API. **Removed the duplicate; net −25 lines.**

**No `same_origin`, no new route.** `upload_url` is now nullable, and null is the whole signal:

```
POST /api/runs/<runid>/generate-upload-url
  → { upload_url: "https://storage.googleapis.com/…" }   # gcs   → PUT there
  → { upload_url: null }                                 # local → POST to upload-dataset
```

Reporting *absence* is better than what I had for a reason beyond tidiness: `same_origin: true` was the server telling the client "attach your bearer token to this URL." A server should never be in that position — if the value were ever wrong, the client would ship credentials somewhere it shouldn't. Now the client's rule is local and unconditional, and it lives in exactly one module (`gcsUpload.ts` → `datasetUpload.ts`): presigned URL gets a bare PUT, null gets a multipart POST with the auth header. Callers just forward `upload_url`; `useRunSetup` lost its branch and its `authBridge` import.

Still XHR rather than the existing `RunsApi.uploadDataset`, because that goes through `BaseApi` → `fetch`, and only XHR reports upload progress — the run-setup UI displays bytes/percent/ETA.

**Two pre-existing gaps in `upload-dataset` that this makes load-bearing**, both fixed here:

- **It enforced no size limit at all** — `@requires_auth()` with no `check_permissions` and no `content_length` check. The validation in `generate-upload-url` is on a *client-asserted* `file_size_bytes`, so it was never sufficient on its own. The 50/100GB cap is now applied where the bytes actually arrive, with the same `HIGHER_UPLOAD_LIMIT` permission.
- **It buffered the whole file to a temp file** (`file.save(tmp.name)`) and then re-uploaded it — a second full copy of a multi-GB upload on the API container's disk. Now streams `file.stream` straight into the store.

Also added the run-exists check (previously a nonexistent run would surface as a 500) and rejection of filenames containing path separators. Filenames are otherwise stored **verbatim, not `secure_filename`d** — run metadata records dataset names and the job looks the files up by them, so normalizing would desync the two.

**Verified against the running stack:** `upload_url: null` as expected with no `same_origin` field; multipart POST lands at the right key; the removed PUT route 404s; `../../escape.csv` → 400 with nothing written outside the run; unknown run → 404; and a submitted job still reads the API-uploaded dataset through its bind mount. 236 tests pass, UI `tsc` + lint clean.

One loose end I noticed but did not touch: `RunsApi.uploadDataset` and the `runs/actions.ts` `uploadDataset` server action both target this same route and **neither is called anywhere**. Part of the answer to "why were there two" is that the client already had unused plumbing for it. Happy to delete both if you want the cleanup in this PR.

dirkraft:
Cleaned up in de1e561 — **−52 lines, pure deletion.** Both dead helpers gone:

- `RunsApi.uploadDataset` + `UploadDatasetResponseBody`
- `runs/actions.ts` `uploadDataset` + `UploadDatasetResponse`

`datasetUpload.ts` is now the only client-side path to the route. `tsc`, `next lint`, and `yarn build` all clean.

---

While confirming they were dead I turned up an **unrelated, pre-existing dead branch** in the same area. Not touched — flagging rather than fixing, since it's outside this PR and deleting a component is your call:

- `ui/src/app/runs/components/RunConfiguration.tsx` is **imported by nothing**.
- It is the **only** consumer of `runs/actions.ts` (`saveMetadata`, `submitRun`), so that server-actions file is effectively dead too.
- `actions.ts`'s `saveMetadata` posts to `POST /api/runs/metadata`, **which doesn't exist** — the real route is `POST /api/runs/<runid>/metadata`. It's presumably never fired because the only caller is unreachable, i.e. the file has drifted out of sync with the API and nothing noticed.

So the honest full answer to "why did we have two in the first place" is that the client had three ways to reach this endpoint, two of them unused and one of them pointing at a route that no longer exists. Two are now gone; say the word if you want `RunConfiguration.tsx` + `actions.ts` looked at in a separate PR.

api/runs/runs_api.py
```diff
@@ -676,9 +690,13 @@ def delete_run(runid: str):
             return jsonify({"error": str(e)}), 500
 
     @api.route("/upload-dataset", methods=["POST"])
-    @requires_auth()
+    @requires_auth(check_permissions=[PermissionType.HIGHER_UPLOAD_LIMIT])
```
dirkraft, line 693: calling this api doesn't require the higher upload limit. one can still upload files. why is the permission added here?

dirkraft:
Pushed 5e97a70 — you're right, `null` was still making the client *interpret* the response instead of just executing it.

**Why it was null, though** (not arbitrary): the two destinations differ in more than the URL — `PUT` raw bytes vs `POST` multipart with a `runid` field, plus credentials on one and not the other. So returning a URL uniformly wouldn't have made the *client* uniform; it would have hidden the branch behind a field that looks uniform but isn't, with the client still carrying hardcoded knowledge of our endpoint's protocol.

The fix is to send the rest of what's needed, so the response describes a **complete request**:

```jsonc
// gcs
{ "upload_url": "https://storage.googleapis.com/…?X-Goog-Signature=…",
  "upload_method": "PUT",  "upload_fields": null }

// local
{ "upload_url": "/api/runs/upload-dataset",
  "upload_method": "POST", "upload_fields": { "runid": "…" } }
```

The client performs what it's handed — `fields` present ⇒ multipart with the file appended, absent ⇒ raw body:

```ts
xhr.open(target.method, url.toString());
if (authToken) xhr.setRequestHeader('Authorization', `Bearer ${authToken}`);
if (target.fields) { /* FormData: fields, then file */ } else { /* raw body */ }
```

No sentinel, no per-backend knowledge, and the endpoint path is gone from the client. `useRunSetup` no longer passes `runid` to the uploader either — it travels in `fields`.

Two things worth noting:

- **This is the shape a third backend would need anyway.** S3's `generate_presigned_post` returns exactly a url plus form fields, so a presigned-POST store drops in without touching the client. It isn't speculative generality — it's the same shape as the case we already have.
- **The store interface is unchanged.** `signed_upload_url()` still returns "a URL you can PUT to" or `None` (GCS, S3, and Azure SAS all support presigned PUT). The descriptor lives purely at the wire layer, which is where the two protocols actually have to be reconciled.

Credentials remain a **client-side** decision — token attached only when the URL resolves to our own origin — rather than something the server requests. That's the one part of the earlier `same_origin` design I'd argue was wrong on the merits, not just on tidiness.

**Verified both branches:** drove a real upload against the running stack using nothing but the returned descriptor (file landed at the right key), and confirmed a mocked presigned result yields `PUT` with `upload_fields: null`. 236 tests, `tsc`/lint/build clean.

For the record, this field went through three shapes before landing: a dedicated streaming route → `same_origin` boolean → nullable `upload_url` → descriptor. The design doc's "Deltas from the proposal" now records that trail so the next reader doesn't re-derive it.

api/runs/models.py
```diff
@@ -336,10 +336,29 @@ class GenerateUploadUrlRequestModel(BaseModel):
...
+    upload_fields: dict[str, str] | None = Field(
+        None,
+        description=(
+            "When set, send the upload as multipart/form-data containing these fields "
+            "plus the file under the field name 'file'. When null, send the raw file as "
+            "the request body. (Presigned PUT URLs use the latter; presigned POST "
+            "policies and this API's own endpoint use the former.)"
+        ),
+    )
+    gcs_path: str = Field(..., description="Storage URI where the file will be stored")
```
dirkraft: If this is not necessarily a "GCS" path, rename it more generically.

api/runs/runs_api.py
```diff
@@ -116,33 +115,47 @@
...
+        rewrite_token = None
+        while True:
+            rewrite_token, bytes_rewritten, total_bytes = dest_blob.rewrite(
+                source_blob, token=rewrite_token
+            )
+            if rewrite_token is None:
+                break
+    else:
+        with source_blob.open("rb") as stream:
+            dest_store.write_stream(dest_key, stream)
```
dirkraft: we went through the trouble of abstracting an ObjectStore, so let's stick to an object-oriented approach?

docs/design/storage-backends.md
```diff
@@ -0,0 +1,274 @@
...
+  nullable `upload_url` — were dropped in review: `upload-dataset` already did the job, and
+  a sentinel still left the client interpreting instead of just executing.
+- The in-container mount point stays `/mnt/gcs` even for the local backend. The deployed
+  Cloud Run job definition pins that path (`--add-volume-mount` in
+  `rebuild_and_deploy.sh`), so renaming it would couple this change to a Cloud Run
+  redeploy. It is now the single constant `backends.base.JOB_MOUNT_ROOT`, so renaming it
+  later is a one-line change plus a redeploy.
+- The Asta workspace handoff (`asta_gcs.py`) still writes to a GCS bucket — that bucket is
+  Asta's, not ours — but now reads its source through the store, doing a server-side copy
+  only when the source is also GCS.
```
dirkraft: verify the end state described by this documentation as of this PR is still accurate

packages/autodiscovery_jobs/src/autodiscovery_jobs/backends/base.py
```diff
@@ -17,12 +17,21 @@
...
 # "process"/"local" run it inside the job container (subprocess / in-process).
 _CODE_EXECUTION_BACKENDS = ("process", "local", "modal")
 
+#: In-container path where the run's persisted data is mounted. Historically named
+#: for GCS because Cloud Run mounts the bucket there, and the deployed Cloud Run
+#: job definition pins it (``--add-volume-mount`` in
+#: ``packages/autodiscovery/scripts/rebuild_and_deploy.sh``). Every backend
+#: reproduces the same path so job arguments stay backend-agnostic, whether the
+#: mount comes from GCS FUSE or a host bind mount.
+JOB_MOUNT_ROOT = "/mnt/gcs"
```
dirkraft: we do not need to maintain historical naming of "gcs" which could imply gcs even when in local storage mode. Please change

packages/autodiscovery_jobs/src/autodiscovery_jobs/backends/docker.py
```diff
@@ -82,27 +92,39 @@ def run_job(self, userid: str, jobid: str, **kwargs) -> str:
...
+        if needs_fuse:
+            # The container gcsfuse-mounts the bucket itself in its entrypoint.
+            environment["GCSFUSE_BUCKET"] = self.config.bucket
+            environment["GCSFUSE_ONLY_DIR"] = prefix
+            environment["GOOGLE_APPLICATION_CREDENTIALS"] = _CONTAINER_GCP_KEY_PATH
+            # Bind-mount the GCP credentials file so gcsfuse (and google-cloud
+            # clients) can authenticate. In docker-out-of-docker the bind source
+            # must be a *host* path, so it is provided out-of-band via
+            # GCP_KEY_HOST_PATH (the same file docker-compose binds into the API).
+            host_key_path = os.environ.get("GCP_KEY_HOST_PATH")
```
dirkraft: .get? this is required at this point, right?

packages/autodiscovery_jobs/src/autodiscovery_jobs/gcs.py
```diff
@@ -1,1373 +1,90 @@
...
+    "purge_user_data",
+    "read_experiment_node",
+    "read_rich_outputs",
+    "soft_delete_job",
+    "summarize_user_data",
+    "upload_dataset",
+    "upload_job_args",
+    "upload_metadata",
+    "write_shared_run_index",
+]
```
dirkraft: do we need this shim file?

packages/autodiscovery_jobs/src/autodiscovery_jobs/exceptions.py
```diff
@@ -19,8 +19,25 @@ class JobAlreadyExistsError(AutodiscoveryJobError):
...
-    """Raised when GCS operations fail."""
+class StorageError(AutodiscoveryJobError):
+    """Raised when a persistence (object store) operation fails."""
+
+    pass
+
+
+# Historical name from when persistence was GCS-only. Retained as an alias so
+# existing `except GCSError` / import sites keep working across both backends.
+GCSError = StorageError
```
dirkraft: don't maintain GCS name. update import sites with new name

scripts/send_completion_emails.py
```diff
@@ -57,21 +58,44 @@ class LockError(Exception):
...
+    from google.cloud import storage
+
+    if config.storage_backend != "gcs":
+        raise LockError(
+            f"--acquire-lock requires STORAGE_BACKEND=gcs (got {config.storage_backend!r}): "
+            "the lock relies on a GCS generation precondition for atomic create. Run "
+            "without --acquire-lock if concurrent invocations are not possible."
+        )
+
+    return storage.Client(project=config.project_id).bucket(config.bucket).blob(LOCK_OBJECT_KEY)
```
dirkraft: why don't we add this to the store interface and implement in each class?

api/runs/models.py
dirkraft: Renamed to `storage_path` in 0240fb2e — response model, the endpoint, `RunsApi.ts`, and the `storagePath` state field in `useRunSetup.ts`.

api/runs/runs_api.py
dirkraft: Agreed. 0240fb2e adds `import_from_gs(uri, key)` / `export_to_gs(key, uri)` to `ObjectStore`. `GcsStore` does both as a server-side rewrite; `FilesystemStore` streams. `sync_preloaded_dataset` is now just `dest_store.import_from_gs(...)`, and `asta_gcs.py` (same pattern in the other direction) uses `export_to_gs`. The startup validator is the only remaining `isinstance` check on the store.

docs/design/storage-backends.md
dirkraft: Re-verified end to end in 0240fb2e: the interface listing now includes `create_exclusive` and the GCS transfer methods, the filesystem-parity table has the create-if-absent row back, the file layout drops the shim, and every stale "Deltas" bullet (`GCSError` alias, `gcs_path`, `/mnt/gcs`, the `asta_gcs` special case) is corrected. The remaining `gcsfuse` mentions are history notes describing what was removed.

packages/autodiscovery_jobs/src/autodiscovery_jobs/backends/base.py
dirkraft: Renamed to `/mnt/data` in 0240fb2e. `rebuild_and_deploy.sh`, `run_job.sh`, the Dockerfile comment, tests, and docs are updated. Deploy note: the Cloud Run job definition pins this path, so `rebuild_and_deploy.sh` needs to run when this lands — until then Cloud Run jobs launched by the new API would look for data under a path the old job definition doesn't mount. Called out as callout 4 in the PR body.

packages/autodiscovery_jobs/src/autodiscovery_jobs/gcs.py
dirkraft: No — deleted in 0240fb2e. Nothing in-repo imported it (one docstring in `purge_user_data.py`, updated). With it gone, GitHub also detects `gcs.py` → `persistence.py` as a rename, so the PR diff shrank by roughly 1,800 lines.

packages/autodiscovery_jobs/src/autodiscovery_jobs/exceptions.py
dirkraft: Alias removed in 0240fb2e; `GCSError` is gone. The three sites in `runs_api.py` and the package `__init__` exports use `StorageError`.

scripts/send_completion_emails.py
dirkraft: Done in 0240fb2e: `create_exclusive(key, data) -> bool` is on the interface. `GcsStore` uses the generation-0 precondition; `FilesystemStore` uses `os.open(O_CREAT | O_EXCL)`, deliberately bypassing the staging/rename path since a rename would clobber. The lock code here is now backend-agnostic and the GCS-only refusal is gone. Tests cover single and contended creates on the filesystem and the `if_generation_match=0` kwarg on GCS.

dirkraft:
Tagging a bunch of y'all on this for whatever thoughts/reactions, not looking for line-by-line code review. Maybe just read the PR description and see if that much makes sense.

My primary goal is to eliminate dependency on gcp/gcs. With this PR it becomes possible to run the full stack including web UI without any cloud volume dependency -- e.g. deploy the AutoDiscovery stack to a VM. One could also provide their own implementation of the ObjectStore interface against whatever persistence backend of choice.

As of this PR, the only remaining external dependency is the LLM provider. This is most immediately valuable to partner work specifically interested in private operation of the AutoDiscovery web stack and the jobs it launches.

PR #59, opened 2026-08-05
Add GitHub Copilot provider support

SuspiciousIdea:
## Summary
- Adds GitHub Copilot as a configurable LLM provider.
- Adds provider credential handling, model configuration, and focused provider tests.

## Validation
- Provider behavior validated through the local desktop flow and focused tests.

## Stack
Base layer for the local desktop runtime stack.


packages/autodiscovery/src/autodiscovery/agents.py
```diff
@@ -141,6 +145,48 @@ def _analyze_image(self, image_data: str) -> str:
         Returns:
             Analysis text
         """
+        if self.llm_provider == "copilot":
```
dirkraft: If copilot is the vision model i'd rather this be keyed on self.vision_model

docs/autodiscovery/standalone.md
```diff
@@ -26,6 +26,52 @@ export VERTEX_ACCESS_TOKEN=$(gcloud auth print-access-token)
...
+   --vision_model claude-haiku-4.5 \
+   --embedding_provider copilot \
+   --embedding_model text-embedding-3-small \
+   --embedding_dimensions 1536 \
+   --backend process \
+   --dedupe \
+   data/measurements.csv
+```
+
+`--llm_provider current` and `--embedding_provider current` are the defaults and
```
dirkraft: rather than create a new value called "current" can we leave the defaults in place if no parameter is specified? Said differently, I'd rather not provide any meaning for "current".

dirkraft:
Thanks for the PRs! I've started going through this first one. 

My usual review style is to comment as I notice things so it might be a sporadic stream of notifications. You can respond or push further commits specific to this PR whenever you like. If the comment notifications are annoying, I can stage them and submit them all as one review instead. Let me know if there's a style you prefer.

packages/autodiscovery/src/autodiscovery/agents.py
```diff
@@ -104,11 +106,13 @@ def __init__(
             backend: Async sandbox executor (ModalEphemeralExecutor or _ProcessBackendAdapter)
             timeout: Timeout in seconds (for Autogen compatibility)
             vision_model: Model to use for image analysis
+            llm_provider: LLM provider used for image analysis.
             usage_tracker: Optional usage tracker for image analysis calls.
         """
         self._executor = backend
         self._timeout = timeout
         self.vision_model = vision_model
+        self.llm_provider = llm_provider
```
dirkraft, line 115: is this additional parameter necessary? If possible, I think I'd rather try to key behavior solely off the existing vision_model parameter 

For switching behavior you can do something similar to `is_gemini_model(self.vision_model)` that appears elsewhere in this file

dirkraft:
I will say, I am flip flopping a bit on how to make specifying different llm providers and models more consistent. I might open a debt ticket rather than complicate this PR and the ones stacked on it. Want to think on it more. The litellm package _may_ help consolidate calls into various models and their respective providers.

dirkraft (approved):
I have a TODO ticket #67 to come back around and make selection of models and providers more consistent via litellm. So happy to take this as is. Feel free to merge if you don't intend to make any other changes on this PR.

dirkraft:
I went ahead and took this now because I want to pursue the litellm transformation of how models are called throughout the codebase.

PR #68, opened 2026-08-21
Route provider/model selection and transport through litellm

### LINKED ISSUE #67: Consistent LLM provider/model selection via litellm naming conventions
> **Updated 2026-08-19** to track `copilot-provider` @ `d7a03eed` ("Address Copilot
> provider review feedback"). That commit removed the `"current"` sentinel in
> favor of `llm_provider: str | None = None`, so the earlier version of this
> ticket's `'current' | 'copilot'` framing no longer describes #59. See
> [Sequencing](#sequencing) for what changed.

## Problem

Provider/model selection in `packages/autodiscovery` is spread across several
mechanisms that don't agree with each other:

1. **Implicit prefix routing.** `get_openai_config` (`agents.py:595`) picks a
   provider by string-sniffing the model name: `is_gemini_model()`
   (`utils.py:17`) is `model.split("/")[-1].startswith("gemini")` → Vertex,
   `model.startswith("o")` → OpenAI o-series handling, everything else → OpenAI.
2. **An optional provider flag.** `--llm_provider copilot` and
   `--embedding_provider copilot` (added in #59) select a namespace for the model
   flags rather than naming a model. Both default to `None`, meaning "use
   model-name routing".
3. **A transport detail promoted to user-facing config.** `--model` values are
   normalized to `google/<model>` for Vertex's OpenAI-compatible endpoint
   (`utils.py:29`), so `google/` appears in user-facing model names even though it
   is a wire-format requirement.

Consequences today:

- **Signature defaults contradict the CLI.** `get_agents(model_name="o4-mini")`
  and `vision_model="gpt-4o"` (`agents.py:663,674`), plus `run()`'s `"gpt-4o"`
  defaults (`run.py:296,329`), while `args.py:15,21,27` defaults to
  `gemini-3.1-pro-preview` / `gemini-3-flash-preview` (same in `easy.py:193-195`).
  Reading the function signatures tells you the wrong thing about what a run
  actually uses.
- **`--llm_provider copilot` with default model flags cannot work.** Neither
  `gemini-3.1-pro-preview` nor `gemini-3-flash-preview` exists in Copilot's
  catalog (Copilot has `gemini-2.5-pro` and `gemini-3-pro-preview`). All three of
  `--model`, `--belief_model`, `--vision_model` must be overridden — as
  `docs/autodiscovery/standalone.md:52-54` does — or the run fails on its first
  model call. Nothing validates this at startup.
- **No capability validation.** `--vision_model` is not checked for vision
  support. `copilot.py` already collects `supports.vision` per model in `doctor`
  but nothing consumes it.
- **One flag, three roles.** A single `--llm_provider` drives chat, belief, and
  vision, so mixed-provider runs (e.g. Copilot chat + Vertex vision) are not
  expressible. `--embedding_provider` is already independent, which shows
  per-role selection is the right shape.
- **Per-call-site provider branches.** `if llm_provider == "copilot"` appears in
  `agents.py` (chat config and vision), `utils.py` (`query_llm`), and
  `deduplication.py` (embeddings). Each new provider multiplies these.

## Proposal: route model selection through litellm's naming convention

litellm is **already a dependency of this monorepo** — `packages/agents` pins
`litellm==1.85.0` and uses the convention today:

```python
# packages/agents/src/agents/experiment_agents.py:28
DEFAULT_MODEL: LiteLlm = LiteLlm(model=os.getenv(MODEL_ENV_VAR, "openai/gpt-5-mini"))
```

`packages/autodiscovery` is the outlier. litellm's convention is
`<provider>/<model>` with snake_case provider slugs, from a 149-member
`LlmProviders` enum (1.97.0): `openai/`, `anthropic/`, `vertex_ai/`, `gemini/`,
`azure/`, `bedrock/`, `github_copilot/`, `ollama/`, …

Verified against litellm 1.97.0 via `get_llm_provider()`:

| Input | Resolves to |
|---|---|
| `o4-mini` | `openai` |
| `gemini-3.1-pro-preview` | `vertex_ai` |
| `vertex_ai/gemini-3.1-pro-preview` | `vertex_ai` |
| `gemini/gemini-3.1-pro-preview` | `gemini` (AI Studio — distinct from Vertex) |
| `google/gemini-3.1-pro-preview` | **error** — not a litellm namespace |
| `claude-haiku-4.5` | `anthropic` (was an error in 1.85.0) |
| `github_copilot/claude-haiku-4.5` | `github_copilot` |

Two findings worth calling out:

- **`google/` is not a litellm provider prefix** and errors outright. Under this
  convention it becomes `vertex_ai/gemini-...`, with `google/` demoted back to an
  internal transport detail.
- **Bare `claude-haiku-4.5` resolves to `anthropic`, not Copilot.** That is the
  exact model name in `docs/autodiscovery/standalone.md:52-54`, so the documented
  Copilot invocation names a model that a litellm-based router would send to
  Anthropic direct. Prefixes remove that ambiguity; a separate provider flag does
  not.

litellm 1.97.0 also ships a **static, offline, no-auth registry of 33
`github_copilot/*` models** in `models_by_provider` / `model_cost`, with
`supports_vision`, `supports_function_calling`, and `max_output_tokens` per
model. That is enough to do startup validation of all model flags — including the
vision-capability check — without a network call.

### Sketch

- Accept `<provider>/<model>` in `--model`, `--belief_model`, `--vision_model`,
  `--embedding_model`, resolved via `litellm.get_llm_provider()`.
- Keep bare names working as aliases (litellm resolves `o4-mini` → `openai` and
  `gemini-3.1-pro-preview` → `vertex_ai` on its own), so existing deployment
  configs keep working unchanged.
- Treat `google/<model>` as a legacy alias for `vertex_ai/<model>` at the
  boundary; keep the `google/` rewrite internal to the Vertex transport.
- Validate every model flag at startup against litellm's registry, checking
  `supports_vision` for the vision model.
- Remove `--llm_provider` / `--embedding_provider` once the prefix carries the
  information. Because both are already optional with a `None` default, this is a
  clean removal rather than a value migration — or they can be kept as a
  litellm-style `custom_llm_provider` escape hatch for endpoints a prefix can't
  express.
- Collapse the per-call-site `if llm_provider == "copilot"` branches
  (`agents.py`, `utils.py`, `deduplication.py`) into one resolver.
- Align the stale in-code defaults with the CLI defaults so signatures stop
  lying.

### Adopt litellm naming, not necessarily litellm transport

These are separable, and only the first is proposed here. litellm's
`github_copilot` provider performs GitHub **device-flow auth inside
`get_llm_provider()`** — it printed device codes and blocked for three 60s
attempts during testing. #59's approach (shell out to an already-authenticated
`copilot` CLI via `github-copilot-sdk`) is likely the better transport for a
batch research pipeline. Similarly, Vertex currently goes through its
OpenAI-compatible endpoint with a token refresher, which works and is load-bearing
for existing deployments.

Scope this as: **litellm as the naming/registry/validation authority**, with
existing transports kept. Swapping transports to `litellm.completion()` is a
separate decision.

## Hard constraint

Existing runtime environment configurations must keep working unchanged —
specifically `asta-autodiscovery-private` deployments. Bare model names in
existing configs must continue to resolve to the same providers they do today.
Any migration is additive, with `google/`-prefixed values retained as aliases.

<a name="sequencing"></a>
## Sequencing

**This is not a blocker for #59**, which is being merged. Verified against
`d7a03eed`:

- `llm_provider` / `embedding_provider` default to `None` throughout, so the
  existing OpenAI/Vertex path is untouched when the flags are omitted.
- All new parameters are keyword-with-default, and every call site of the changed
  signatures (`query_llm`, `get_embedding`, `ModalSandboxExecutor`, `get_belief`,
  `save_nodes`, `save_nodes_to_json`, `dedupe`) passes by keyword — no positional
  shifts.
- `github-copilot-sdk` is an optional extra (`[project.optional-dependencies]
  copilot`) with function-local imports, so the default install and import graph
  are unchanged.
- `make test` green: 172 passed, 1 deselected.

`d7a03eed` also replaced the original `"current"` sentinel with `None`
(`--llm_provider` is now `choices=["copilot"], default=None`). That resolved the
main design objection at the source and makes this refactor cheaper — there is no
`"current"` string to migrate out of configs or stored metadata.

### Carry-over for #61

**#61 ("Add local desktop UI and provider onboarding") is now inconsistent with
merged #59** and needs a fix independent of this refactor. It has not been
updated since 2026-08-05, before `d7a03eed`, and still hardcodes
`'current' | 'copilot'` in ~11 places: TypeScript union types, run-request
payloads, persisted-metadata comparisons (`metadata.llmProvider === 'current'`),
and an API provider-list entry `{"id": "current"}`.

That is a concrete break, not just stale naming. #60 puts `llm_provider` in
`optional_params`, filtered only for `None` and empty strings:

```python
# Filter out None and empty strings, but allow 0 and other valid values
optional_params = ["model", "belief_model", "vision_model", "llm_provider", ...]
```

So a UI-supplied `"current"` survives the filter and becomes
`--llm_provider=current`, which fails `choices=["copilot"]` in argparse. The UI's
default provider selection would break every local run.

Fix in #61: drop `'current'` from the unions and omit the field (or send `null`)
for default routing. #60's own model layer is already
`llm_provider: str | None = Field(None, ...)` and is consistent with merged #59 —
only #61's API/UI layer is stale.

Doing that also keeps this refactor cheap, by not re-cementing the provider
concept into the persisted API contract as a closed enum.

### Also still open from #59 (non-blocking)

- No startup validation that model flags exist in Copilot's catalog:
  `--llm_provider copilot` with the default `--model gemini-3.1-pro-preview`
  fails mid-run.
- `github-copilot-sdk` is in the root `dev` dependency-group, so every dev
  environment installs it despite it being an optional extra for users.

dirkraft:
## TL;DR

Model selection now routes through **litellm**, which changes how you name models: **every model flag must be litellm-qualified as `<provider>/<model>`.** Bare model names (and the old `google/` wire prefix) are now rejected at startup with an error naming the correct form. The `--llm_provider` and `--embedding_provider` flags are **removed** — the provider travels with each flag instead, so mixed-provider runs are expressible for the first time.

**Gemini** — now under the `vertex_ai/` provider:

| | |
|---|---|
| Before | `--model gemini-3.7-flash` (provider inferred from the `gemini` name prefix; `google/…` also leaked in) |
| After | `--model vertex_ai/gemini-3.7-flash` |

**GitHub Copilot** — now the `github_copilot/` provider. Interactively, litellm runs GitHub's device-code login on first use and caches the token itself, so no setup is needed. For non-interactive runs, pre-seed a token at `$GITHUB_COPILOT_TOKEN_DIR/access-token` (default `~/.config/litellm/github_copilot`) — litellm does no TTY detection, so a headless run without one blocks on a device prompt for ~3 minutes before failing:

| | |
|---|---|
| Before | `--llm_provider copilot --model claude-haiku-4.5 --embedding_provider copilot` |
| After | `--model github_copilot/gpt-4.1 --embedding_model github_copilot/text-embedding-3-small` |

> Example uses `gpt-4.1` rather than `claude-haiku-4.5`: Copilot's Claude models silently ignore `response_format` and return plain text, which fails parsing since every agent and belief call passes a pydantic schema. `gpt-4.1` and `gpt-5.4` are verified working end to end.

**Also breaking, beyond naming:** Vertex no longer reads `VERTEX_ACCESS_TOKEN` (it uses ADC — `GOOGLE_APPLICATION_CREDENTIALS` or `gcloud auth application-default login`; no deployment sets it today), and Copilot no longer shells out to the `copilot` CLI, so the `[copilot]` extra and `python -m autodiscovery.copilot doctor` are gone.

More examples: `--model openai/o4-mini`, `--vision_model vertex_ai/gemini-3.7-flash`. Unqualified or non-litellm names fail fast at startup (e.g. `gemini-3.7-flash` → use `vertex_ai/gemini-3.7-flash`), and `github_copilot/<model>` names are checked against the *account's own* `/models` endpoint (litellm's static Copilot catalog is wrong in both directions). That endpoint is still a superset of what chat will serve, so it narrows the failure window rather than closing it.

---

Closes #67.

## What

Provider/model selection in `packages/autodiscovery` was spread across three mechanisms that didn't agree: model-name prefix sniffing, an optional `--llm_provider` flag, and a `google/` wire-format prefix that leaked into user-facing config. On top of that, each provider had its own hand-written client.

This routes everything through litellm — **both** the `<provider>/<model>` naming convention **and** the transport. `autodiscovery/llm.py` is now the only file in the package that knows a provider exists; every other module passes a model string through.

> The issue scoped transport out ("adopt litellm naming, not necessarily litellm transport"). That was revisited in review: both stated blockers turned out tractable, and keeping the transports was what made the naming-only version *add* code instead of removing it.

## Provider layer, before and after

Deleted outright:

| File | Lines |
|---|---|
| `copilot_provider.py` — Copilot CLI SDK runtime + AG2 adapter | 512 |
| `copilot.py` — Copilot `doctor` diagnostics | 137 |
| `vertex_client.py` — Vertex client + OAuth token refresher | 95 |
| `vertex_config.py` — Vertex base-URL derivation | 41 |
| `llm_retry.py` — OpenAI-SDK and AG2 `OpenAIWrapper` monkeypatches | 155 |

Replaced by one file: `llm.py`, 314 lines of which **99 are code** (the rest docstrings, per the repo's ruff `D` convention).

## Name-based routing removed

| Before | After |
|---|---|
| `is_gemini_model()` — `startswith("gemini")` | the `vertex_ai/` prefix |
| `is_reasoning_model()` — `startswith("o")`/`"gpt-5"` | `litellm.drop_params` |
| `model_name.startswith("o")` in `get_openai_config` | `litellm.drop_params` |
| `normalize_vertex_model_name()` → `google/` | litellm owns Vertex's wire format |
| `if llm_provider == "copilot"` × 4 | gone — one transport for every provider |
| duplicated `_is_gemini_model` in the injected local-backend patch | patch calls `autodiscovery.llm` directly |
| `get_openai_config` + `CopilotAG2Client` | one `LiteLLMAG2Client` |
| AG2 usage via an `OpenAIWrapper` monkeypatch | recorded in `LiteLLMAG2Client.create` |

AG2 0.10 picks a client by `api_type` from a fixed provider list; registering our own means AG2 inherits litellm's list instead.

## What's still ours, and why

Three model facts litellm's registry cannot express — each verified, not assumed:

- **`accepts_temperature`** — `get_supported_openai_params("o4-mini")` *includes* `temperature`, so `drop_params` won't strip it and the call would fail at the API.
- **`max_n`** — no registry field for a provider's per-request sample cap (Vertex 5, OpenAI reasoning 8).
- **`normalize_reasoning_effort`** — `minimal` → `low` where unsupported. This one is now driven by litellm's per-model `supports_minimal_reasoning_effort` rather than a `gpt-5` name prefix, which **fixes a latent bug**: the old rule downgraded the whole gpt-5 family, and a test asserted that wrong behaviour.

Plus prefix validation and startup checks.

## Model names

Every model flag takes litellm's `<provider>/<model>`. The prefix is **required** — a bare name is ambiguous (`claude-haiku-4.5` is Anthropic direct or Copilot depending on who you ask), and resolving one means calling `litellm.get_llm_provider()`, which authenticates (see below).

```sh
--model github_copilot/claude-haiku-4.5 --vision_model vertex_ai/gemini-3.1-pro-preview
```

Mixed-provider runs are expressible for the first time, since the provider travels with each flag. `--llm_provider` and `--embedding_provider` are removed.

Unqualified and non-litellm names are rejected at startup, naming the form to use:

```
'gemini-3.1-pro-preview' is missing a provider. Model names are litellm-qualified
as <provider>/<model>, e.g. vertex_ai/gemini-3.1-pro-preview or openai/gemini-3.1-pro-preview.
```

`google/…` is rejected the same way — it was Vertex's OpenAI-compat wire prefix, never a litellm provider.

## Startup validation

`run.resolve_model_args` checks every flag against litellm's offline registry before the first model call: the vision model must support image input (the check #67 notes nothing consumed), the embedding model must be an embedding model, chat flags must be chat models, and Copilot models must exist in Copilot's catalog. That last one catches the #59 failure mode directly:

```
--model=github_copilot/gemini-3.1-pro-preview is not in github_copilot's catalog.
Available: claude-haiku-4.5, claude-opus-4.5, ..., gemini-3-pro-preview, ...
```

Unmapped models are a *warning* for OpenAI and Vertex, which ship models faster than litellm maps them, and an *error* for Copilot, whose catalog litellm ships as a complete enumeration.

## Breaking changes

Both documented in `standalone.md`, `configuration.md` and `RELEASE.md`.

**1. Vertex auth.** `VERTEX_ACCESS_TOKEN`, `GOOGLE_OAUTH_ACCESS_TOKEN` and `VERTEX_OPENAI_BASE_URL` are no longer read. litellm's Vertex client uses Application Default Credentials and refreshes on its own; there is no raw-bearer-token entry point. Deploy config uses `GOOGLE_APPLICATION_CREDENTIALS`, so `asta-autodiscovery-private` should be unaffected — **this is the claim worth confirming**, since it's #67's hard constraint. Local dev moves to `gcloud auth application-default login`.

**2. Copilot auth.** No longer shells out to the `copilot` CLI via `github-copilot-sdk`. litellm reads a GitHub OAuth token from `$GITHUB_COPILOT_TOKEN_DIR/access-token`. The `[copilot]` extra and `python -m autodiscovery.copilot doctor` are removed — which also resolves #67's note about `github-copilot-sdk` sitting in the root `dev` group.

**3. Model flags** must be prefixed, as above. In-repo configs are updated (`args.py`, `easy.py`, every in-code default, `scripts/run_job.sh`, `autodiscovery_jobs` docstrings); external configs need updating too.

## litellm and Copilot device-flow auth

Worth recording, because it constrains any future work here: `litellm.get_llm_provider()` — and the `litellm.supports_*()` helpers that call it — run GitHub's device flow when handed a `github_copilot/` model, printing a device code and blocking for three 60s attempts. Confirmed on 1.85.0 and 1.97.0. There is **no env var to disable it**; the `GITHUB_COPILOT_*` variables only relocate token files and endpoints.

`llm.py` therefore splits the prefix itself and only reads the static registry via `get_model_info(custom_llm_provider=...)`, which is offline and auth-free. A regression test stubs `Authenticator` to raise and fails if anything reintroduces an authenticating call.

## Dependency

`litellm` 1.85.0 → `>=1.97.0,<2`, added to `packages/autodiscovery` (it was only a `packages/agents` dependency; a uv workspace resolves one version, so the `agents` pin moved too). `ag2[openai, gemini]` → `ag2[openai]`, and `github-copilot-sdk` is dropped from both the optional extra and the root `dev` group.

## Testing

`make test`: **177 passed**, 1 deselected (172 on `main`). `pyright` on the package: 62 → 58 errors, all pre-existing; `llm.py` is clean.

**Not exercised:** no live API call has been made from this branch. The Vertex ADC path in particular deserves one real run before merge.

## Not in scope

- The `api/metrics/costs.py` hand-maintained price table — split out as #70.
- #67's #61 carry-over (`'current'` in the UI/TS layer) — those files don't exist on `main` yet.


dirkraft:
## Can the interactive login be disabled by env var?

**No.** Checked litellm 1.97.0's `llms/github_copilot/authenticator.py` — every `GITHUB_COPILOT_*` variable it reads is a path, URL, or client-id override:

`GITHUB_COPILOT_TOKEN_DIR`, `GITHUB_COPILOT_ACCESS_TOKEN_FILE`, `GITHUB_COPILOT_API_KEY_FILE`, `GITHUB_COPILOT_API_KEY_URL`, `GITHUB_COPILOT_DEVICE_CODE_URL`, `GITHUB_COPILOT_ACCESS_TOKEN_URL`, `GITHUB_COPILOT_CLIENT_ID`, `GITHUB_COPILOT_API_BASE`

There's no non-interactive / headless flag. The device flow is unconditional in `get_access_token()` whenever the token file is missing or empty, and it's on the resolution path, not just the request path:

```
get_llm_provider("github_copilot/x")
  -> GithubCopilotConfig()._get_openai_compatible_provider_info()
     -> Authenticator.get_api_key()
        -> get_access_token()
           -> _login()          # prints a device code, blocks 3 x 60s
```

### There is a non-interactive workaround

`get_access_token()` reads the token file first and returns immediately if it's non-empty — `_login()` is never reached. So a deployment can mount a GitHub OAuth token and point litellm at it:

```sh
export GITHUB_COPILOT_TOKEN_DIR=/var/run/secrets/copilot   # contains: access-token
```

Verified with a deliberately invalid token — it fails fast instead of hanging:

```
$ GITHUB_COPILOT_TOKEN_DIR=/tmp/ghcp python -c 'litellm.get_llm_provider("github_copilot/claude-haiku-4.5")'
401 Unauthorized for https://api.github.com/copilot_internal/v2/token   (x3)
raised BadRequestError: ... Failed to refresh API key
elapsed 0.3s   (device flow would be ~180s)
```

Caveat: it's non-interactive but not offline — it still calls `api.github.com/copilot_internal/v2/token` unless the cached `api-key.json` is unexpired.

### For this PR it's moot, and now enforced

`model_spec` never calls `get_llm_provider()` or `supports_*()` for prefixed models, so nothing on the deployed path can reach the authenticator. The exposure was only ever a future change — someone adding a `supports_vision()` call, or swapping to `litellm.completion()` transport.

Rather than rely on that staying true, f7b7d700 adds a regression test that stubs `Authenticator.get_api_key` / `get_access_token` / `_login` to raise, then resolves and validates a Copilot model. It fails if anything reintroduces an authenticating litellm call. The finding and the `GITHUB_COPILOT_TOKEN_DIR` escape hatch are documented in the `model_spec` module docstring.

`make test`: 197 passed.

dirkraft:
## When does the device flow actually fire?

**Only when the provider resolves to `github_copilot`.** Instrumented `Authenticator` to raise instead of block, then walked every path:

| Call | Auth fired? |
|---|---|
| `import litellm` | no |
| `get_llm_provider("gpt-4o")` → openai | no |
| `get_llm_provider("gemini-3.1-pro-preview")` → vertex_ai | no |
| `get_llm_provider("vertex_ai/gemini-3.1-pro-preview")` | no |
| `get_llm_provider("openai/o4-mini")` | no |
| `get_llm_provider("github_copilot/gpt-4o")` | **YES** |
| `get_model_info(model=..., custom_llm_provider="openai")` | no |
| `get_model_info(model=..., custom_llm_provider="github_copilot")` | no |
| `supports_vision(model=..., custom_llm_provider="github_copilot")` | **YES** |

So it is not an import-time or startup cost — an OpenAI/Vertex deployment never touches it, which is why the existing default path was never at risk.

Two details worth recording:

**The entry point is `get_api_base()`, not `get_api_key()`.** `GithubCopilotConfig._get_openai_compatible_provider_info` calls `authenticator.get_api_base()` first, but that one is benign — it reads the cached `api-key.json` and returns `None` on failure. The blocking call is `get_api_key()` on the very next line, which falls through to `_refresh_api_key()` → `get_access_token()` → `_login()`.

**`supports_vision()` fires even with an explicit `custom_llm_provider`.** With auth stubbed out it silently returned `False` for a model that does support vision — so it fails open, quietly. Another reason the registry-only `get_model_info()` path is the right one.

### Bare names can't route into it — but that wasn't guaranteed

`parse_model` hands unprefixed names to `get_llm_provider()`, so a bare name that resolved to `github_copilot` would authenticate. I checked the names most likely to do it — the ones that exist *only* in Copilot's catalog:

```
mai-code-1-flash       -> LLM Provider NOT provided   (no auth)
gpt-41-copilot         -> LLM Provider NOT provided   (no auth)
claude-opus-4.6-fast   -> LLM Provider NOT provided   (no auth)
gpt-4-o-preview        -> LLM Provider NOT provided   (no auth)
gpt-5.3-codex          -> openai                      (no auth)
```

None route to Copilot, so the bare-name path is safe today. But that's litellm's behavior, not a guarantee — a registry change upstream could turn it into a hang. 097cb2ec extends the guard to the bare-name path (parameterized over the defaults plus those Copilot-only names), so that regression surfaces as a test failure instead of a stuck deployment.

`make test`: 206 passed.

dirkraft:
## Is `model_spec.py` necessary?

Fair question — I measured instead of defending it.

**Size after trimming:** 313 lines, of which **129 are code** (105 docstrings, mandated by the repo's ruff `D` convention; 16 comments; 63 blank). 8dd3a7ea cut the dead surface: `known_models()` had zero callers outside one error message, and `is_known`/`mode` were used only inside `validate_model`.

**What of those 129 lines litellm could supply: almost none of the hard parts.**

| Concern | In litellm? |
|---|---|
| Split prefix, pass `custom_llm_provider` explicitly | No — and it's the whole reason we don't call `get_llm_provider()`, which auths for Copilot |
| `google/` Vertex wire prefix | Not a litellm concept at all |
| `accepts_temperature` | **No.** `supported_openai_params` lists `temperature` for `o4-mini`, which rejects it — checked |
| `max_n` (Gemini 5, o-series 8) | No such field |
| `google`/`copilot` aliases | Ours |
| Fallback for models litellm hasn't mapped | Ours by definition |
| Supported-provider gate + closed-catalog check | Ours |
| Caching | `get_model_info` is uncached and sits on the `query_llm` hot path |

Roughly 95 of the 129 lines are logic litellm cannot provide. It has to live *somewhere*; the alternative is scattering it back across `utils.py`, `agents.py`, and `deduplication.py`, which is the problem #67 opens with.

**The design that would actually delete it** is adopting litellm's *transport* (`litellm.completion()`) — then routing, temperature, and reasoning-effort quirks are litellm's problem. The issue deliberately scoped that out (Vertex's token-refresher path is load-bearing, and litellm's Copilot transport does device-flow auth). So this module is the cost of "litellm as naming authority, existing transports kept." If we ever revisit the transport decision, `model_spec` is what goes away.

### Auditing it turned up a real bug

Checking whether litellm could replace the last two hand-rolled rules surfaced `supports_minimal_reasoning_effort`. The old rule keyed off a `gpt-5` name prefix:

```python
def is_reasoning_model(model):
    return any(model.startswith(p) for p in ["o", "gpt-5"])
```

so it downgraded `minimal` → `low` for the entire gpt-5 family — but gpt-5 models *do* accept `minimal` effort, and litellm records that per model. There was a test asserting the wrong behavior:

```python
model="gpt-5-mini", reasoning_effort="minimal"
assert request["reasoning_effort"] == "low"     # wrong
```

efee5cab fixes it: downgrade only when the registry doesn't affirm support. o-series still maps to `low`, Gemini and Copilot still pass through, and the corrected test now covers both. I retargeted the old test at `o4-mini` (where the assertion is correct) and added a gpt-5 case — flagging it because it changes previously-asserted behavior.

Note the registry only ever says `True` or nothing, never `False`, so a missing value means "unknown" — the property is treated as a positive signal only, which is why the rule stays gated on `is_openai`.

`make test`: 208 passed.

dirkraft:
## Transport adopted — the PR is now net −903 src lines

You were right that the additive LoC was diagnostic. It was: the naming-only version reimplemented provider knowledge in a 326-line module instead of delegating it. With the transport swapped, the module mostly stops carrying that weight and the plumbing it was wrapping disappears.

### What was the reasoning for not doing it initially?

The issue's own **"Adopt litellm naming, not necessarily litellm transport"** section scoped it out — device-flow auth in litellm's Copilot provider, and Vertex's OpenAI-compatible endpoint being "load-bearing for existing deployments." I took that as a given rather than re-testing it. That was the wrong call: both concerns turned out to be tractable, and I should have checked before accepting the constraint.

### Deleted outright

| File | Lines |
|---|---|
| `copilot_provider.py` — Copilot CLI SDK runtime + AG2 adapter | 512 |
| `llm_retry.py` — OpenAI-SDK / AG2 `OpenAIWrapper` monkeypatches | 153 |
| `copilot.py` — Copilot doctor diagnostics | 137 |
| `vertex_client.py` — Vertex client + token refresher | 95 |
| `vertex_config.py` — Vertex base-URL derivation | 41 |

### Collapsed

- `get_openai_config` + `CopilotAG2Client` → one `LiteLLMAG2Client`. AG2 0.10 picks a client by `api_type` from a fixed provider list; registering our own means AG2 inherits litellm's list instead.
- `query_llm`, `get_embedding` and vision each lost their provider branch and client construction.
- The injected local-backend matplotlib patch had its *own* copy of Vertex auth and model-name logic. That backend runs in-process, so the patch now imports `autodiscovery.llm` directly.
- AG2 usage tracking moved from a monkeypatch into `LiteLLMAG2Client.create` — the one point every response flows through.
- `litellm.drop_params` handles `reasoning_effort` per model, so that gating left `model_spec` and every call site. `wire_model_name` is gone too; litellm owns Vertex's `google/` prefix.

### What `model_spec` still carries, and why

Two rules litellm genuinely cannot express, both verified rather than assumed:

- **`accepts_temperature`** — `get_supported_openai_params("o4-mini")` *includes* `temperature`, so `drop_params` will not save us; the call would fail at the API.
- **`max_n`** — no registry field for a provider's per-request `n` cap.

Plus name resolution, aliases, and startup validation. Everything else went.

### Two breaking changes — please sanity-check the first

**Vertex auth.** `VERTEX_ACCESS_TOKEN`, `GOOGLE_OAUTH_ACCESS_TOKEN` and `VERTEX_OPENAI_BASE_URL` are no longer read. litellm's Vertex client uses ADC and refreshes itself, and there is no raw-bearer-token entry point. I checked that deploy config sets `GOOGLE_APPLICATION_CREDENTIALS`, not `VERTEX_ACCESS_TOKEN`, so `asta-autodiscovery-private` should be unaffected — but that's the one claim worth you confirming, since it's the hard constraint. Local dev moves to `gcloud auth application-default login`.

**Copilot auth.** No longer shells out to the `copilot` CLI. litellm reads a GitHub OAuth token from `$GITHUB_COPILOT_TOKEN_DIR/access-token`. The `[copilot]` extra and `python -m autodiscovery.copilot doctor` are removed, which also resolves the issue's note about `github-copilot-sdk` sitting in the root `dev` group.

Both are documented in `standalone.md`, `configuration.md` and `RELEASE.md`.

### Verification

`make test`: 184 passed (was 172 on main; the drop from 208 is the deleted Copilot-SDK and monkeypatch suites, replaced by `LiteLLMAG2Client` tests). `pyright`: 62 → 59, all pre-existing. `llm.py` and `model_spec.py` are clean.

Not exercised: no live API call against Vertex or Copilot has been made from this branch. The Vertex ADC path in particular deserves one real run before merge.

dirkraft:
## Aliases dropped — and yes, it deleted more than the aliases themselves

Answering both questions directly.

### Do we have logic that exists *because of* not-quite-litellm names?

Yes, more than the aliases. Everything below existed only to support unqualified or non-litellm names, and all of it is now gone:

| Removed | Why it existed |
|---|---|
| `_legacy_provider()` — the `gemini*` → `vertex_ai` heuristic | litellm can't resolve a bare name it hasn't mapped yet, so a new preview model would have broken configs on a stale registry. **This was the last name-prefix test in the codebase.** |
| The bare-name path through `litellm.get_llm_provider()` | Needed to turn a bare name into a provider. It was also **the only call in the package that could trigger Copilot's device-flow auth** — so that entire risk class is gone, along with the guard test we added for it. |
| `_PROVIDER_ALIASES` / `canonical_provider()` | `google/` (Vertex's OpenAI-compat wire prefix, never a litellm provider) and `copilot` → `github_copilot`. |
| `--llm_provider` / `--embedding_provider`, the `default_provider` parameter threaded through `parse_model` and `validate_model`, their deprecation warnings, `PROVIDER_FLAG_HELP` | The pre-prefix way to say which provider a bare name meant. |
| `_DEFAULT_COPILOT_EMBEDDING_MODEL` | Only needed because `--embedding_provider copilot` combined with a bare default (`text-embedding-3-large`) that Copilot has no equivalent for. |

`resolve_model_args` is now four `validate_model` calls and no longer rewrites `args` — the flags already hold canonical names.

`model_spec` is down to **108 code lines** (137 naming-only → 129 after the trim → 108 now).

### Further negative commits?

**Src is now −908 lines** vs main. The alias removal itself is roughly break-even in raw count (deleted logic, added the qualified names back into defaults and tests), but it removes the two things that were still genuinely wonky: the name-prefix heuristic and the auth-triggering resolver call.

Configurations updated to litellm names: `args.py` and `easy.py` defaults, every in-code default, `scripts/run_job.sh`, `autodiscovery_jobs` docstrings, and the docs. Unqualified names now fail at startup naming the form to use:

```
'gemini-3.1-pro-preview' is missing a provider. Model names are litellm-qualified
as <provider>/<model>, e.g. vertex_ai/gemini-3.1-pro-preview or openai/gemini-3.1-pro-preview.
```

`google/…` is rejected too, with the `vertex_ai/` suggestion.

### One more piece of not-quite-litellm logic, deliberately left alone

`api/metrics/costs.py` has a **26-model hand-maintained price table** plus a `_lookup_pricing` that strips provider prefixes — with a comment referencing `google/` specifically. litellm's registry carries `input_cost_per_token`/`output_cost_per_token` for every model, and with this PR litellm also computes `response_cost` per call (already surfaced through `LiteLLMAG2Client.get_usage`), so that table is redundant in principle.

I left it because `api/` is a separate lightweight Flask service — its own `requirements.txt`, pinned to Python 3.11, no litellm — and pulling litellm into a metrics dashboard for a price lookup is a poor trade. The prefix-stripping still works correctly for both old (`google/…`) and new (`vertex_ai/…`) stored run data, since it splits on `/`. Worth a separate issue if you want the table gone; say the word and I'll open one.

### Verification

`make test`: 176 passed. `pyright`: 62 → 58, all pre-existing. Startup behaviour spot-checked across qualified, unqualified, `google/`, mixed-provider, bad-vision and off-catalog-Copilot cases.

dirkraft:
Filed #70 for the `api/metrics/costs.py` price table so this PR stays scoped to provider/model routing. Nothing further is going into #68 for it.

dirkraft:
## `model_spec.py` is deleted

You were right to keep asking. I tested it properly this time instead of defending it, and it doesn't survive.

`ModelSpec` was a frozen dataclass whose `__str__` returned exactly the string it was constructed from. Both of its original reasons had already been removed by earlier commits in this PR:

- **`wire_model_name`** existed to translate a name for Vertex's OpenAI-compatible endpoint (`google/gemini-…`). litellm owns provider wire formats now.
- **`provider`** had to be *derived*, by asking litellm to resolve a bare name. Requiring a prefix put the provider in the string.

What was left was a parse step that reconstructed its own input. The consumer census made that obvious:

| Field | Consumers in `src/` |
|---|---|
| `is_copilot` | **0** |
| `is_vertex` | 1 |
| `is_openai` | 1 |
| `supports_vision` | 0 (only inside `validate_model`) |
| `provider` | 4 — all `model.split("/")[0]` |
| `str(spec)` | round-trips the input |

So models are now plain `<provider>/<model>` strings end to end — which is what litellm takes anyway — and the handful of facts that are genuinely ours moved into `llm.py`, next to the transport that needs them:

```python
provider_of(model)                     # split + validate the prefix
model_info(model)                      # cached, offline registry lookup
accepts_temperature(model)             # litellm wrongly claims o4-mini accepts it
max_n(model)                           # no registry field for an n cap
normalize_reasoning_effort(model, e)   # minimal -> low where unsupported
validate(model, flag=..., ...)         # startup checks
```

`normalize_reasoning_effort` also moved out of `utils.py` — it's a model rule, not a utility. `ModelSpecError` became `llm.ModelError`.

**`llm.py` is 99 code lines and is now the only file in the package that knows a provider exists.** Everything else passes a model string through.

### Where the PR landed

| Stage | Src delta |
|---|---|
| Naming only (first version) | **+346** |
| + litellm transport | −903 |
| + required prefixes | −908 |
| + `model_spec.py` deleted | **−1018** |

The whole provider layer that existed before: `copilot_provider.py` (512), `copilot.py` (137), `vertex_client.py` (95), `vertex_config.py` (41), and 153 lines of monkeypatches in `llm_retry.py` — all gone, replaced by one 314-line file (99 code, the rest docstrings under the repo's `D` convention).

`make test`: 177 passed. `pyright`: 62 → 58, all pre-existing; `llm.py` clean. Startup behaviour re-verified across qualified, unqualified, `google/`, mixed-provider, non-vision and off-catalog-Copilot cases.

packages/autodiscovery/src/autodiscovery/agents.py
```diff
@@ -96,46 +89,23 @@ def __init__(
...
         usage_tracker: UsageTracker | None = None,
     ):
         """Initialize the sandbox executor wrapper.
 
         Args:
             backend: Async sandbox executor (ModalEphemeralExecutor or _ProcessBackendAdapter)
             timeout: Timeout in seconds (for Autogen compatibility)
-            vision_model: Model to use for image analysis
-            llm_provider: LLM provider used for image analysis.
+            vision_model: Vision model, optionally qualified as ``<provider>/<model>``
```
dirkraft: when is the / notation optional?

packages/autodiscovery/src/autodiscovery/agents.py
```diff
@@ -592,76 +445,98 @@ def get_logs(
         return "SimpleCodeBlockTransform", True
 
 
-def get_openai_config(
-    api_key: str | None = None,
+class LiteLLMAG2Client:
```
dirkraft, line 450: a quick google claims that ag2 supports litellm. Can you look into this? Do we need this custom class?

packages/autodiscovery/src/autodiscovery/deduplication.py
```diff
@@ -300,10 +240,8 @@ def dedupe(
     merge_threshold=0.7,
     seed=42,
     rep_mode="biggest",
-    model="gpt-4o",
-    llm_provider=None,
-    embedding_provider=None,
-    embedding_model=None,
+    model="vertex_ai/gemini-3-flash-preview",
+    embedding_model="openai/text-embedding-3-large",
```
dirkraft: is this the embedding model used on autodiscovery.allen.ai?

packages/autodiscovery/src/autodiscovery/llm.py
```diff
@@ -0,0 +1,314 @@
...
+    """Import and configure litellm once.
+
+    Returns:
+        The configured ``litellm`` module.
+    """
+    import litellm
+
+    # Let litellm strip parameters a given model does not accept (for example
+    # reasoning_effort on gpt-4o) instead of hand-maintaining that per call site.
+    litellm.drop_params = True
```
dirkraft, line 76: what's the alternative?

packages/autodiscovery/src/autodiscovery/llm.py
```diff
@@ -0,0 +1,314 @@
...
+import os
+from typing import Any
+
+from autodiscovery.llm_retry import load_retry_config
+
+# litellm provider slugs this package has credentials and defaults for.
+OPENAI = "openai"
+VERTEX_AI = "vertex_ai"
+GITHUB_COPILOT = "github_copilot"
+SUPPORTED_PROVIDERS = (OPENAI, VERTEX_AI, GITHUB_COPILOT)
```
dirkraft: why are we constrained to this list of supported providers, rather than everything that litellm recognizes?

packages/autodiscovery/pyproject.toml
```diff
@@ -12,22 +12,19 @@ dependencies = [
...
     "tenacity>=9.1.2",
+    "litellm>=1.97.0,<2",
     "asta-autodiscovery-modal",
     "asta-code-execution",
 ]
 
-[project.optional-dependencies]
-copilot = [
-    "github-copilot-sdk>=1.0.7,<2",
-]
```
dirkraft, line 30: What dependencies do I need to call openai, gemini, copilot, etc... models? Is there a published, preferred way in the context of litellm?

packages/autodiscovery/RELEASE.md
```diff
@@ -90,35 +90,42 @@ export GOOGLE_APPLICATION_CREDENTIALS=/path/to/service-account-key.json
...
-| Flag | What it controls | Provider |
+| Flag | What it controls | Default |
 | --- | --- | --- |
-| `--model` | Primary reasoning model used for hypothesis generation and analysis. | Gemini if name starts with `gemini`, else OpenAI. |
-| `--belief_model` | Model used for belief updates over experimental outcomes. | Same routing. |
-| `--vision_model` | Model used to interpret plots and figures emitted by experiments. | Same routing. |
+| `--model` | Primary reasoning model used for hypothesis generation and analysis. | `vertex_ai/gemini-3.1-pro-preview` |
+| `--belief_model` | Model used for belief updates over experimental outcomes. | `vertex_ai/gemini-3-flash-preview` |
+| `--vision_model` | Model used to interpret plots and figures emitted by experiments. | `vertex_ai/gemini-3.1-pro-preview` |
+| `--embedding_model` | Model used for deduplication embeddings. | `openai/text-embedding-3-large` |
```
dirkraft, line 125: want to be sure this default matches the deployment of autodiscovery.allen.ai since that is essentially the hosted demo of autodiscovery.

packages/autodiscovery/src/autodiscovery/agents.py
dirkraft: It is not — that docstring was stale. The prefix has been required since d28d0c16 ("Require litellm-qualified model names"); this docstring and 16 others were left behind.

Fixed in 620fc30c: they all now read "as litellm's `<provider>/<model>`". Unqualified names are rejected at startup:

```
'gemini-3.1-pro-preview' is missing a provider. Model names are litellm-qualified
as <provider>/<model>, e.g. vertex_ai/gemini-3.1-pro-preview or openai/gemini-3.1-pro-preview.
```

Good catch — thanks.

packages/autodiscovery/src/autodiscovery/llm.py
dirkraft: Fair challenge; the allow-list was doing less than it looked like. Removed in 620fc30c.

What it actually bought: a startup error for a real-but-unconfigured provider (`anthropic/…`) instead of an auth failure on the first model call. What it cost: capping us at 3 of litellm's ~149 providers for no structural reason — supplying credentials is the operator's job either way, and the provider-specific bits already no-op for providers they don't name (`_provider_kwargs` only maps Vertex's env vars; `accepts_temperature` / `max_n` return the neutral default). That is exactly the artificial restriction this PR is supposed to be removing.

`provider_of` still validates the slug against litellm's `LlmProviders`, so a typo is still a startup error:

```
'vertexai' in 'vertexai/gemini-3.1-pro-preview' is not a litellm provider.
See https://docs.litellm.ai/docs/providers for the full list.
```

`vertex_ai`, `openai` and `github_copilot` remain the three we document credentials for and test; anything else works if you supply its env vars.

packages/autodiscovery/src/autodiscovery/llm.py
dirkraft: Failing the request. With `drop_params = False` (litellm's default) litellm forwards every parameter as given, and the provider rejects the ones it does not accept — e.g. `reasoning_effort` on `gpt-4o` comes back as a 400 mid-run.

Before this PR that was handled by hand, per call site: `is_reasoning_model()` gated `reasoning_effort`, and `get_openai_config` gated `logprobs` on `model_name.startswith("o")`. `drop_params` replaces all of that with litellm's per-model parameter table.

The one place it is not sufficient is `accepts_temperature`, three functions below — `get_supported_openai_params("o4-mini")` *includes* `temperature`, so litellm would not strip it and the call fails at the API. That is why that rule is still ours, and the docstring says so.

packages/autodiscovery/src/autodiscovery/agents.py
dirkraft: Looked into it. AG2 0.10 has **no** in-process litellm client, so yes, we still need this — but the google result is not wrong either, it just means something different.

What is actually in the package: `autogen/interop/litellm/litellm_config_factory.py`. Despite the name it is not a model client — it translates an AG2 `llm_config` into the shape **crawl4ai** wants (crawl4ai uses litellm internally for web crawling). Nothing to do with routing AG2's own calls.

What AG2 dispatches on, from `autogen/oai/client.py`:

```
anthropic  azure  bedrock  cerebras  cohere  google  groq  mistral  ollama  responses  together
```

No `litellm`. And the declared extras (`ag2[anthropic]`, `ag2[gemini]`, `ag2[ollama]`, …) contain no litellm entry.

So "AG2 supports litellm" in practice means **the litellm proxy server**: you run `litellm --config …` as a separate OpenAI-compatible HTTP process, then point AG2 at it with `api_type: openai` + `base_url`. Real and documented, but for a batch pipeline it means a second process or container per job to start, supervise and fail independently.

`register_model_client` is AG2's sanctioned alternative — its own docstring says "a custom client class that follows the ModelClient interface". `LiteLLMAG2Client` is **39 code lines** and replaced two things (`get_openai_config` plus the old `CopilotAG2Client`), so it is a net reduction, not an addition.

Happy to switch to the proxy if you would rather not carry the class, but I would call that a downgrade for this workload. Leaving as-is unless you disagree.

packages/autodiscovery/src/autodiscovery/deduplication.py
dirkraft: Yes — and worth flagging that it is unchanged, not newly chosen: `main` had `model or "text-embedding-3-large"` in `get_embedding`, so this is the same model, now provider-qualified.

Whether autodiscovery.allen.ai ever reaches it is a separate matter. `build_job_args` never passes `--embedding_model`, so it always falls back to this default; and embeddings only run under `--dedupe`, which the job I looked at had off (`'dedupe': False`).

One thing to be aware of if dedupe is ever turned on there: chat runs on Vertex but this embedding model is OpenAI, so the deployment would need `OPENAI_API_KEY` set alongside the Vertex credentials. That was equally true before this PR — the prefix just makes it visible.

packages/autodiscovery/RELEASE.md
dirkraft: **It does not, and chasing this turned up a blocker — see below.**

On the table itself: these are the CLI defaults. `--vision_model` and `--embedding_model` are never passed by `build_job_args`, so the deployment does use those two defaults. `--model` and `--belief_model` are overridden per run by the API, so for those the table describes the CLI, not the hosted demo.

### The blocker

`ui/src/app/runs/components/RunConfiguration.tsx` sends **bare** model names:

```ts
const MODEL_OPTIONS = [
    { value: 'gpt-4o', label: 'GPT-4o' },
    { value: 'o4-mini', label: 'o4-mini' },
    { value: 'gpt-5-mini', label: 'GPT-5 Mini' },
    { value: 'gpt-5-nano', label: 'GPT-5 Nano' },
];
const [model, setModel] = useState('gpt-4o');
```

Those flow through `submitRun({model, belief_model})` into `build_job_args` as `--model=gpt-4o --belief_model=gpt-4o`. After this PR that is rejected before the first model call:

```
'gpt-4o' is missing a provider. Model names are litellm-qualified as <provider>/<model>,
e.g. vertex_ai/gpt-4o or openai/gpt-4o.
```

So **merging this as-is breaks every run launched from autodiscovery.allen.ai.** It fails fast and loudly rather than silently, but it fails.

Two more things that fall out of the same file, unrelated to the prefix change:

1. The dropdowns offer only OpenAI models, while the CLI defaults and the deployment's Vertex credentials are Gemini. So the UI already cannot select the models the deployment is provisioned for.
2. `MODEL_OPTIONS` and `BELIEF_MODES` are identical four-entry literals, duplicated.

I have not touched `ui/` in this PR. How do you want to handle it?

- **(a)** Fix it here — prefix the four values (`openai/gpt-4o`, …), which is a small diff but pulls `ui/` into this PR.
- **(b)** Separate PR, and mark this one blocked-by so they land together.
- **(c)** I open an issue and you sequence it.

I would lean **(a)**: the two changes are only safe together, and a stacked PR that must merge simultaneously is more coordination than the diff is worth. Your call.

packages/autodiscovery/pyproject.toml
dirkraft: Checked this properly. **litellm pulls the OpenAI SDK itself**, so the answer is mostly "nothing extra":

```
litellm -> openai>=2.20.0, httpx, tiktoken, tokenizers, pydantic, aiohttp, jinja2, jsonschema, ...
```

Per provider:

| Provider | Extra dependency | Credentials |
| --- | --- | --- |
| `openai` | none — comes with litellm | `OPENAI_API_KEY` |
| `vertex_ai` | `google-auth` (for ADC) | `GOOGLE_APPLICATION_CREDENTIALS`, or `gcloud auth application-default login` |
| `github_copilot` | none — plain HTTP | GitHub OAuth token at `$GITHUB_COPILOT_TOKEN_DIR/access-token` |

On the published/preferred way: litellm deliberately does **not** ship per-provider extras. Its own extras are `[proxy]`, `[caching]`, `[utils]` and similar — infrastructure, not providers. Providers are expected to work off the base install plus whatever SDK a given one needs (Vertex being the notable case, because Google auth is not pure HTTP). So there is no `litellm[vertex]` to point at.

This did surface one real cleanup, in 620fc30c: **`google-cloud-aiplatform` is gone.** It was only ever needed by `vertex_client.py`, which this PR deletes, and is now imported nowhere in the repo. It was also how `google-auth` was arriving — transitively. Replaced with an explicit `google-auth` and a comment saying why. That drops a large SDK (and its `google-cloud-*` tail) from the install.

I left `ag2[openai]==0.10` alone. The extra is redundant now that litellm brings `openai`, and we no longer use AG2's OpenAI client, but AG2 internals import `openai` in places I have not fully traced — not a change I would make without a real run behind it. Worth a follow-up if you want the install trimmed further.

dirkraft:
Went through all seven comments — replies are inline. Summary:

**Changed** (620fc30c):
- **17 stale docstrings** still said model names are "optionally qualified". They've been required since d28d0c16; the docstrings were left behind.
- **Dropped the `SUPPORTED_PROVIDERS` allow-list.** It capped us at 3 of litellm's ~149 providers for no structural benefit — the provider-specific bits already no-op for providers they don't name. The slug is still validated against litellm's `LlmProviders`, so a typo remains a startup error.
- **`google-cloud-aiplatform` → `google-auth`.** aiplatform was only needed by the deleted `vertex_client.py` and is now imported nowhere; it was also how `google-auth` arrived, transitively. Drops a large SDK and its `google-cloud-*` tail from the install.

**Answered, no change:**
- **AG2 + litellm**: AG2 0.10 has no in-process litellm client. `autogen/interop/litellm/` is a *crawl4ai* config translator, and no `api_type` dispatches to litellm. "AG2 supports litellm" means the litellm **proxy server** — a second process per job. `LiteLLMAG2Client` is 39 code lines and replaced two things. Happy to reconsider if you disagree.
- **`drop_params`**: the alternative is a 400 mid-run. It replaces the old hand-rolled `is_reasoning_model()` / `startswith("o")` gating.
- **Embedding model**: unchanged from `main`, just prefixed. Never overridden by the job launcher, and only reached under `--dedupe` (off in the job I inspected).

## One blocker found

Chasing the RELEASE.md question turned up a real problem: **`ui/` sends bare model names**, so merging this as-is breaks every run launched from autodiscovery.allen.ai.

```ts
// ui/src/app/runs/components/RunConfiguration.tsx
const MODEL_OPTIONS = [{ value: 'gpt-4o', ... }, { value: 'o4-mini', ... }, ...];
const [model, setModel] = useState('gpt-4o');
```

These reach the job as `--model=gpt-4o`, which is now rejected at startup. It fails fast and loudly, not silently — but it fails.

Also worth noting from the same file: the dropdowns offer **only OpenAI models**, while the CLI defaults and the deployment's Vertex credentials are Gemini — so the UI already can't select what the deployment is provisioned for. And `MODEL_OPTIONS`/`BELIEF_MODES` are duplicate literals.

I haven't touched `ui/`. Options in the inline reply; I'd lean toward fixing it in this PR, since the two changes are only safe together. **Your call — I'll hold until you say.**

`make test` 177 passed; pyright 58, all pre-existing.

packages/autodiscovery/RELEASE.md
dirkraft: **Correction: this is not a blocker.** I traced the UI value only as far as `submitRun` before; following it the rest of the way, it never reaches the job.

`/api/runs/submit` accepts only `{runid, userid}`:

```python
class SubmitRunRequestModel(BaseModel):
    runid: str
    userid: str
```

`model` and `belief_model` arrive in the POST body and are dropped. The handler then rebuilds config from `metadata.json`, and its allow-list omits both:

```python
optional_params = [
    "exploration_weight", "mcts_selection", "surprisal_width",
    "evidence_weight", "warmstart_experiments", "n_warmstart",
]
```

The UI does not put them in metadata either — it saves `{title, datasets}`. So `manager.run_job(model=None, belief_model=None)` every time, `build_job_args` skips both flags, and the job falls back to the `args.py` defaults baked into the image.

Confirmed against a real run:

```
'model': 'gemini-3.1-pro-preview', 'belief_model': 'gemini-3-flash-preview'
```

Those are the CLI defaults, not the `gpt-4o` the dropdown had selected. So bare names never reach the prefix validation, and this PR does not break autodiscovery.allen.ai. Sorry for the false alarm.

### What it does mean

To your question on the other thread — you are right that model selection is effectively **stack-level**, not per-request. It lives in the job image's `args.py` defaults. The per-request plumbing exists (`manager.run_job(model=...)` → `build_job_args`) but nothing wires the API to it.

So the UI dropdowns are dead code, and actively misleading twice over: they advertise a choice with no effect, and they offer only OpenAI models while the deployment is provisioned for Vertex. Removing them is the honest fix; if a per-stack override is wanted later, an env var would match the `ASTA_AGENTS_MODEL` precedent. Still not touching `ui/` in this PR.

dirkraft:
**Correction to my last comment: the UI is not a blocker.** I'd traced the model value only as far as `submitRun`. Following it further, it never reaches the job.

`/api/runs/submit` accepts only `{runid, userid}` — `model`/`belief_model` are dropped at the boundary — and the handler's `optional_params` allow-list omits both when rebuilding config from metadata. The UI doesn't put them in metadata either (it saves `{title, datasets}`). So `run_job(model=None, belief_model=None)` every time, and the job uses the `args.py` defaults baked into the image.

Verified against a real run: `'model': 'gemini-3.1-pro-preview'`, not the `gpt-4o` the dropdown had selected.

So bare names never reach the prefix validation. **This PR doesn't break autodiscovery.allen.ai**, and needs no `ui/` change to merge.

What it does mean: model selection is effectively **stack-level** today, living in the job image's CLI defaults. The per-request plumbing exists (`manager.run_job(model=...)` → `build_job_args`) but the API never wires into it — so the UI dropdowns are dead code that advertise a choice with no effect, and offer only OpenAI models while the deployment runs on Vertex. Worth removing, separately from this PR.

dirkraft:
## Took the dead UI out (7a8a937f)

Agreed on the general practice — pulling this in only because it's pure subtraction with no import to update and no behaviour to preserve, and because it's the thing that misled me into calling this PR a blocker.

**Deleted:**
- `ui/src/app/runs/components/RunConfiguration.tsx` (−255) — never imported, in any revision. Created in `98c8895d` (2026-01-15, "allow users to start runs") *alongside* `RunSetup.tsx`; two run-setup components landed in one commit and only one got wired up. That commit is the only one in `ui/` history to mention the name at all.
- `ui/src/app/runs/actions.ts` (−158) — its only importer was the file above. It's a superseded duplicate of `api/RunsApi.ts`, which already has `saveMetadata`, `uploadDataset`, `submitRun`, `deleteRun` — the live `useRunSetup → RunsApi` path uses those. Its other exports (`RunDetails`, …) resolve from `@/types/Run` everywhere else.

**Fixed:** `RunSetup.tsx`'s header comment claimed the component configures "models". It never did — its only `Select` is `mcts_selection`. That was the sole stale model reference in the surviving files; `useRunSetup.ts` and `RunParametersModal.tsx` mention models nowhere.

### On the provenance, since it explains the confusion

The eight commits touching `RunConfiguration.tsx` after creation are maintenance-by-accident — `"Fix lint"`, `"Fix abs imports"`, `"lint"`, plus incidental edits from a credits-field rename and a sidebar-title change. Its OpenAI-only dropdown (`gpt-4o`, `o4-mini`, `gpt-5-mini`, `gpt-5-nano`) came from `d7f434c4`, titled **"Fix lint"**, the same day the file was created. So it was scaffolding a lint pass reflowed, not a considered model list — which is why it disagreed with the Vertex-provisioned deployment.

### Verification

`ui`: `tsc --noEmit` clean, `eslint` clean, `next build` clean (all routes build). `make test`: 177 passed.

Net for the PR is now **−1431** lines across `src/` and `ui/`.

dirkraft:
Some manual tests on cli executed to completion, no fatal errors, no _obvious_ problems.

gemini defaults
```
uv run --package asta-autodiscovery python -m autodiscovery.easy \
 --name "test" \
  --description "seattle pet licenses" \
  --n_experiments 4 \
  --out_dir ./results/ \
  ~/Downloads/Seattle_Pet_Licenses_20260126.csv \
```

and with openai models
```
    --model openai/gpt-4o \
    --belief_model openai/gpt-4o-mini \
    --vision_model openai/gpt-4o
```

and with github_copilot hosted models
```
    --model github_copilot/gpt-4.1 \
    --belief_model github_copilot/gpt-4.1 \
    --vision_model github_copilot/gpt-4.1
```

dirkraft:
## Copilot validation now uses the account's live catalog (04e971ae)

Found by actually running the CLI: `github_copilot/gpt-5` passed our startup validation and then failed on the first real call with

```
litellm.BadRequestError: Github_copilotException - The requested model is not supported.
```

That killed the first MCTS iteration, and since the root node ships one seeded experiment with `allow_generate_experiments=False`, every later iteration printed "No new experiment generated" — 24k iterations in 2s with zero LLM calls.

### The `_CLOSED_CATALOG_PROVIDERS` premise was wrong

I'd assumed litellm ships Copilot's catalog as a complete enumeration. It isn't authoritative in *either* direction. Comparing litellm 1.97's 33 entries against this account's live `/models` (57 entries):

| Model | litellm catalog | actually callable |
|---|---|---|
| `gpt-5` | listed → we **accepted** | **no** |
| `gpt-5.1`, `gpt-5.2`, `gemini-3-pro-preview` | listed | no |
| `gpt-5.4`, `gpt-5.5`, `gpt-5.6-*` | absent → we **rejected** | yes (`n` + `response_format` verified) |
| `claude-opus-5`, `claude-sonnet-5`, `grok-4.6`, `kimi-k3` | absent → we **rejected** | yes |

So it rejected working models and accepted broken ones.

### Fix

`validate()` now prefers Copilot's `/models` endpoint for `github_copilot`, using it for both existence and vision support, and falls back to litellm's registry when unavailable. `_CLOSED_CATALOG_PROVIDERS` is gone — an unmapped model is a warning for every provider now, since the registry is advisory.

Crucially, `_copilot_live_catalog()` reads **only** litellm's cached, unexpired `api-key.json` and never touches `Authenticator`, so it cannot trigger the device-flow login. There's a test for exactly that: token dir pointed at an empty directory with `Authenticator` stubbed to raise, asserting it returns `None` rather than authenticating.

Verified after the change:

```
github_copilot/gpt-5     REJECT  not available to this Copilot account
github_copilot/gpt-5.4   ACCEPT  (warns: not in litellm's registry)
github_copilot/gpt-4.1   ACCEPT
github_copilot/gpt-4o-mini  REJECT  does not support image input  (as --vision_model)
```

### Two Copilot quirks worth recording, not fixed here

Both verified with real API calls:

- **`n` is silently ignored by some models.** `github_copilot/gpt-5-mini` returns 1 choice for `n=5`. So `--n_belief_samples 5` quietly becomes 1 sample. `gpt-4.1` and `gpt-5.4` honour it correctly. Nothing detects this.
- **`response_format` is silently ignored by the Claude models.** `github_copilot/claude-haiku-4.5` with a pydantic schema returns `ok`, not JSON — which would fail `_parse_choices`. litellm's registry has no `supports_response_schema` for them, and Copilot's `/models` reports `structured_outputs: None`.

So for this pipeline the safe Copilot choice today is `github_copilot/gpt-4.1` (vision, schema, and `n` all verified working).

### Run status

Vertex ADC and OpenAI both ran to completion, which closes the "no live API call" caveat for those two paths. Copilot is verified at the call level but hasn't yet completed a full run.

`make test`: 179 passed. `pyright`: clean on `llm.py`.

dirkraft:
Filed #73 to investigate `n` effectiveness against the primary-model set (TBD, blocked on #71 and Copilot-collaborator input). Nothing from it lands here — the Gemini defaults are verified unaffected by the transport swap.

Two things in #73 that touch this PR specifically:

1. **A regression #68 introduces for Copilot only.** On `main`, `copilot_provider` looped client-side and never sent `n`, so any `n_samples` worked. Here Copilot goes through litellm with `n=n_samples` in one call and `max_n` returns `None` for it, so `n > 8` hard-errors (`Invalid 'n': ... Expected a value <= 8`). Latent at the default `n_belief_samples=5`. One-line fix, left out to keep this PR scoped to the Gemini defaults per review direction.
2. **#68 and #71 conflict semantically.** #71 sets the defaults to bare `gemini-3.7-flash`; this PR requires litellm-qualified names and rejects unqualified ones at startup. Whichever merges second needs `vertex_ai/gemini-3.7-flash`. They also overlap textually on `args.py`, `easy.py`, `run.py`, `agents.py`, `scripts/run_job.sh` and `standalone.md`.

dirkraft:
Caught up with `main` in ceccdddf — this now includes #71 (Use Gemini 3.7 Flash).

Five conflicts, all the semantic collision I flagged on both PRs: #71 set the defaults to a bare `gemini-3.7-flash`, which this branch rejects at startup. Resolved by taking #71's model choice with this branch's naming — every default is now `vertex_ai/gemini-3.7-flash`:

| File | Conflict |
| --- | --- |
| `args.py`, `easy.py` | all three model flags |
| `run.py` | `run_mcts` signature defaults |
| `scripts/run_job.sh` | `--model`/`--belief_model`/`--vision_model` |
| `standalone.md` | credentials section |

I also re-aligned the in-code signature defaults that #67 asked to stop lying — `agents.py`, `utils.py`, `beliefs.py`, `mcts_utils.py` and `deduplication.py` still pointed at 3.1-pro / 3-flash-preview, and now follow the CLI. Same for the RELEASE.md defaults table and the standalone.md examples.

Kept from main untouched: the `MessageTokenLimiter(model="gpt-4o")` fix and the `gemini-3.7-flash` pricing entry in `api/metrics/costs.py`.

`make test` 179 passed; pyright 58, all pre-existing; `resolve_model_args` smoke-tested against the new defaults.

One carry-over for #73 rather than here: 3.7-flash is now the default for all three roles, and it's the model with the weakest `candidateCount` independence I measured (5 → 2 distinct, 8 → 3 distinct). Recorded there; no `n` changes in this PR.

packages/autodiscovery/pyproject.toml
dirkraft: Dropped it in 0d2843c2 — you were right that it was redundant.

`ag2`'s `openai` extra pulls `openai>=1.99.3`, and litellm already requires `openai>=2.20.0,<3.0.0` directly. The latter is strictly stronger, so the extra cannot change what gets installed. And since this branch routes every AG2 call through `LiteLLMAG2Client` instead of AG2's own OpenAI client, declaring it was also misleading about *why* `openai` is in the tree.

Verified rather than inferred, since I'd said I wouldn't change it without a real check:

- after `uv sync`, `openai` 2.24.0 is still installed (via litellm)
- `autogen` imports resolve
- `get_llm_config()` returns the expected `LiteLLMAG2Client` entry
- a full `get_agents(..., backend="process")` builds all seven agents with model clients registered and the process backend up
- `make test` 179 passed

So the dependency answer for this PR is now: **litellm brings the OpenAI SDK, Vertex needs `google-auth` for ADC, Copilot needs nothing extra.** That is the whole list — `google-cloud-aiplatform` is gone too (unused after `vertex_client.py` was deleted, and it was only supplying `google-auth` transitively).

dirkraft:
another cli run with the new gemini 3.7 defaults. looks ok at a glance

docs/autodiscovery/standalone.md
```diff
@@ -13,63 +13,126 @@ pip install autodiscovery
...
+# ...or local user credentials:
+gcloud auth application-default login
 ```
 
-To use OpenAI models instead, pass `--model gpt-...` and export `OPENAI_API_KEY`.
+> **Changed:** `VERTEX_ACCESS_TOKEN` is no longer used. litellm's Vertex client
+> authenticates with ADC and refreshes on its own, so a raw access token from
+> `gcloud auth print-access-token` has no equivalent. Use
+> `gcloud auth application-default login` locally; deployments already using
+> `GOOGLE_APPLICATION_CREDENTIALS` need no change.
```
radamson: Nit: I think this might be better in a changelog or just removed.

docs/autodiscovery/standalone.md
```diff
@@ -13,63 +13,126 @@ pip install autodiscovery
...
-If the SDK cannot download its runtime, install the Copilot CLI separately and
-point the SDK at it:
-
-```sh
-export COPILOT_CLI_PATH=/path/to/copilot
-```
+> **Changed:** Copilot no longer shells out to the `copilot` CLI via
+> `github-copilot-sdk`, and the `[copilot]` extra and
+> `python -m autodiscovery.copilot doctor` command are gone. The account still
+> needs an active Copilot seat.
```
radamson: Nit: Should we move to changelog?

packages/autodiscovery/src/autodiscovery/agents.py
```diff
@@ -207,25 +135,19 @@ def _analyze_image(self, image_data: str) -> str:
                 ],
             },
         ]
+        try:
+            response = llm.complete(self.vision_model, messages)
+        except Exception as exc:
+            return f"Image analysis skipped: {exc}"
 
-        response = call_with_backoff(
```
radamson, line 211: Does LiteLLM provide exponential backoff?

packages/autodiscovery/src/autodiscovery/agents.py
```diff
@@ -207,25 +135,19 @@ def _analyze_image(self, image_data: str) -> str:
                 ],
             },
         ]
+        try:
+            response = llm.complete(self.vision_model, messages)
+        except Exception as exc:
+            return f"Image analysis skipped: {exc}"
```
radamson: I think the previous code would have failed here, which would result in retry of the code. If these types of failures are very rare I think this exception handling would be fine, but if its frequent it might causes issues with the result interpretation

(Side Note: the way figures are handled by autodiscovery in general isn't great)

packages/autodiscovery/src/autodiscovery/agents.py
```diff
@@ -375,89 +297,34 @@ def parse_bucket_path(bucket_path: str) -> tuple[str, str]:
 
 
 def build_image_analysis_patch(vision_model: str) -> str:
+    """Build the matplotlib patch injected into locally executed experiment code.
+
+    Only the ``local`` backend uses this, and it runs in-process, so the patch
+    can call straight back into :mod:`autodiscovery.llm` rather than
+    reimplementing provider routing and auth for the execution context.
+    """
```
radamson, line 306: This seems very helpful!

Side Note: Now that we have plots stored with the logs we could probably do something more comprehensive ad always interpret plots outside the exec environment. Definitely better for another PR and could be overcome by a more comprehensive execution change in the works...

radamson (commented):
Using LiteLLM seems great 👍. Also super awesome that this reduces the code surface area by a decent chunk.

I didn't test it, but here were some initial comments from a skim. I don't think any are truly blocking, but we should at least think about the backoff and image analysis changes (if only to keep an eye on them going forward)

packages/autodiscovery/src/autodiscovery/agents.py
dirkraft: Good question — the answer was "not the way we had it configured," so I've changed it back. Fixed in c81c6c1d.

`litellm.num_retries` maps onto the provider SDK's `max_retries`, so the retry *count* survived but the delay shape didn't: `LLM_RETRY_INITIAL_DELAY_SECONDS` and `LLM_RETRY_MAX_DELAY_SECONDS` became inert while still being documented in `docs/configuration.md`. litellm does have a tenacity path, but it's `completion_with_retries()`, it defaults to `retry_strategy="constant_retry"`, and `completion()` doesn't use it.

So `llm.complete` / `llm.embed` now wrap the litellm call in `call_with_backoff` again, with `litellm.num_retries = 0` so there's one retry layer instead of two nested ones. All three env vars are honoured again.

Checked that our predicate still works against litellm's exception types — they subclass openai's, so classification is unchanged:

```
litellm.RateLimitError        retryable=True
litellm.InternalServerError   retryable=True
litellm.APIConnectionError    retryable=True
litellm.Timeout               retryable=True
litellm.BadRequestError       retryable=False
litellm.AuthenticationError   retryable=False
```

Verified live: an auth failure returns in ~2.5s rather than retry-storming, and a real Vertex call still succeeds.

Side effect worth noting — your question caught that `call_with_backoff` had become dead in production code, used only by its own tests. It's load-bearing again.

packages/autodiscovery/src/autodiscovery/agents.py
dirkraft: You're right that behaviour changed, though not quite in the direction described — and I've removed the change in c81c6c1d.

On `main` the exception didn't cause a code retry either: `execute_code_blocks` wrapped `_analyze_image` in its own `try/except` and appended `"Failed to analyze image: ..."` per figure. That handler is still there in this PR, so the `try/except` I'd added inside `_analyze_image` was duplicate handling that shadowed it — same non-fatal outcome, but a different message (`"Image analysis skipped: ..."`) and one that pre-empted the caller.

Now the exception propagates to the existing handler, so the behaviour matches `main` exactly. Left a comment at the call site so the absence of a local `try` reads as deliberate.

Agreed on the side note re: figures generally — and your other comment about interpreting plots outside the exec environment is the more interesting version of that. Happy to open an issue if you want it tracked before the execution rework lands.

docs/autodiscovery/standalone.md
dirkraft: Removed in 8e2898ca — all three `Changed:` blockquotes, since there's no changelog file to move them to and `standalone.md` should just describe the current state.

Nothing is lost: the Vertex ADC requirement is in `RELEASE.md` (the published package readme) and `docs/configuration.md`, and the naming plus Copilot auth changes are now in the PR's TL;DR.

packages/autodiscovery/src/autodiscovery/agents.py
dirkraft: Thanks. And agreed on the side note — with plots already stored alongside the logs, interpreting them outside the exec environment would be strictly better than the in-sandbox matplotlib patch, which only exists for `--backend local`.

Worth noting this PR shrinks that surface rather than growing it: the injected patch used to carry its own duplicated copy of Vertex auth and model-name logic, and now just calls back into `autodiscovery.llm`. So whichever way the execution rework goes, there's less to unpick.

Happy to file an issue for the out-of-environment plot interpretation if you'd like it tracked separately from the execution work.

docs/autodiscovery/standalone.md
dirkraft: Same as the note above — removed in 8e2898ca along with the other two, since there's no changelog to move them to.

dirkraft:
Thanks @radamson — all five addressed. Two code changes came out of it, both reverting things this PR had changed for the worse.

**Backoff (agents.py:211)** — c81c6c1d. Your instinct was right: litellm only partly covers it. `litellm.num_retries` maps onto the provider SDK's `max_retries`, so the retry *count* survived but `LLM_RETRY_INITIAL_DELAY_SECONDS` and `LLM_RETRY_MAX_DELAY_SECONDS` went inert while still documented in `docs/configuration.md`. litellm's tenacity path is `completion_with_retries()`, defaults to `constant_retry`, and isn't what `completion()` uses.

`llm.complete`/`llm.embed` now wrap in `call_with_backoff` again with `litellm.num_retries = 0`. Verified our predicate still classifies litellm's exceptions correctly (they subclass openai's) and that an auth failure fails fast rather than retry-storming. Your question also caught that `call_with_backoff` had gone dead in production, used only by its own tests.

**Image analysis (agents.py:141)** — c81c6c1d. Behaviour did change, though not as described: on `main` the exception didn't trigger a code retry either — `execute_code_blocks` already caught it and appended `"Failed to analyze image: ..."` per figure. That handler is still present, so my inner `try/except` was duplicate handling that shadowed it with a different message. Removed; the exception now reaches the existing handler, matching `main` exactly.

**Doc nits** — 8e2898ca. All three `Changed:` blockquotes removed; no changelog exists to move them to, and the substance is already in `RELEASE.md` and `docs/configuration.md`.

**Also, unprompted:** while checking the TL;DR I corrected three things in it — it recommended `github_copilot/claude-haiku-4.5` as the Copilot example, but Copilot's Claude models silently ignore `response_format` and return plain text, which fails parsing since every agent and belief call passes a pydantic schema. Swapped to `gpt-4.1`, which I've verified end to end. Also clarified that Copilot validation uses the *account's* `/models` endpoint rather than litellm's static catalog, and that naming isn't the only breaking change (Vertex and Copilot auth both moved).

`make test` 179 passed; pyright 57, one below main's baseline.

On your two side notes — figure handling generally, and interpreting plots outside the exec environment — happy to file an issue for the latter if you want it tracked ahead of the execution rework. This PR at least shrinks that surface: the injected matplotlib patch used to carry a duplicated copy of Vertex auth and model-name logic and now just calls into `autodiscovery.llm`.

packages/autodiscovery/src/autodiscovery/agents.py
dirkraft: Filed #74 for this — thanks, it's a sharper framing than I'd had.

Digging into it, the split is worse than "could be more comprehensive": on `--backend local` the vision model is called **from inside the execution environment**. `CodeBlockWrapperTransform` injects a patch that monkeypatches `plt.show`, calls the model in-process, prints the analysis to stdout, and emits token usage as a magic stdout line that the parent scrapes back out. So `local` is simultaneously the least isolated backend and the only one that needs model credentials inside the sandbox.

It also means figures aren't persisted on that path at all — `_get_executor_rich_outputs` returns `[]` because `LocalCommandLineCodeExecutor` has no `get_last_rich_outputs`, so nothing lands in `rich_outputs/` or the report. Which is exactly why your "plots are stored with the logs" observation only holds for `process`/`modal`.

I framed the desired state as one path — sandbox returns figures as data, parent interprets — which deletes `build_image_analysis_patch`, `CodeBlockWrapperTransform`, `LOCAL_IMAGE_USAGE_MARKER` and its parser, and makes analysis independent of the agent calling `plt.show()`.

And I agree it belongs with the execution work: it's really a corollary of #55, since the IPython-kernel-shaped backends are the ones that produce rich outputs. I raised the question there of whether `--backend local` should exist at all — `process` is also local, just isolated, and is the default — in which case this closes by deletion rather than implementation.

PR #71, opened 2026-08-27
Use Gemini 3.7 Flash

radamson:
Swap from Gemini 3.1 Pro to 3.7-Flash ($3.75/1M token output vs $12/1M token output). Also bumps overly conservative token limiter.

### DIFF
--- a/api/metrics/costs.py
+++ b/api/metrics/costs.py
@@ -8,6 +8,7 @@ from datetime import datetime
 # Source: https://cloud.google.com/vertex-ai/generative-ai/pricing
 LLM_PRICING: dict[str, dict[str, float]] = {
     # Google Gemini models
+    "gemini-3.7-flash": {"input": 0.75, "output": 3.75},
     "gemini-3.1-pro-preview": {"input": 2.00, "output": 12.00},
     "gemini-3-pro-preview": {"input": 2.00, "output": 12.00},
     "gemini-3-flash-preview": {"input": 0.50, "output": 3.00},
--- a/docs/autodiscovery/standalone.md
+++ b/docs/autodiscovery/standalone.md
@@ -15,8 +15,7 @@ dependency. Requires Python 3.13+.
 
 ## Credentials
 
-The default models (`gemini-3.1-pro-preview`, `gemini-3-flash-preview`) run on
-Vertex AI. Set:
+The default model (`gemini-3.7-flash`) runs on Vertex AI. Set:
 
 ```sh
 export VERTEX_PROJECT_ID=your-gcp-project
--- a/packages/autodiscovery/README.md
+++ b/packages/autodiscovery/README.md
@@ -34,9 +34,9 @@ uv run --package autodiscovery python -m autodiscovery.run \
     --out_dir="outputs" \
     --dataset_metadata="discoverybench/real/test/nls_ses/metadata.json" \
     --n_experiments=16 \
-    --model="gemini-3-flash-preview" \
-    --belief_model="gemini-3-flash-preview" \
-    --vision_model="gemini-3-flash-preview"
+    --model="gemini-3.7-flash" \
+    --belief_model="gemini-3.7-flash" \
+    --vision_model="gemini-3.7-flash"
 ```
 
 To resume a previous exploration, use the `--continue_from_dir` flag to specify the directory containing the previous
--- a/packages/autodiscovery/scripts/run_job.sh
+++ b/packages/autodiscovery/scripts/run_job.sh
@@ -98,7 +98,7 @@ echo "Executing Cloud Run Job..."
 JOB_ARGS="--dataset_metadata=${METADATA_PATH},--out_dir=${OUTPUT_PATH},--n_experiments=4,--work_dir=work,--use_modal_sandbox,--bucket_path=${BUCKET_PATH},--no-timestamp_dir"
 
 if [ "${USE_GEMINI}" = "true" ]; then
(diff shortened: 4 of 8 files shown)


dirkraft:
any shareable evidence that performance is similar or better? @radamson

radamson:
The main justification is back of the napkin math on cost, release date, and benchmarks (Flash benchmarks higher than Pro on AA's Intelligence Index: https://artificialanalysis.ai/models/comparisons/gemini-3-7-flash-medium-vs-gemini-3-1-pro-preview)

The extent of the testing otherwise was just a small test:

```sh
uv run auto-discovery \
  --name "Tiny smoke test" \
  --description "Tiny smoke-test dataset with two groups and one numeric value column." \
  --n_experiments 1 --batch_size 1 --n_threads 1 \
  --backend process --out_dir smoke/out-flash37 \
  smoke/tiny.csv
  ...
  NODE_LEVEL=2, NODE_IDX=0:
-------------------------

Hypothesis: A point-biserial correlation between binary group indicator and 'value' is strongly positive and statistically significant (r_pb > 0.90).
Prior: 0.5000
Posterior: 0.9545
Surprisal: True
Belief Change: 0.4545 (+)
KL Divergence: 1.2486
Reward: 2.2727
  ```
 
The token usage for this minuscule test was comparable to a previous pro smoke result I had lying around.
 
| run | model | calls | prompt | completion | reasoning | total | est. cost |
|---|---|---|---|---|---|---|---|
| `smoke/pro` | 3.1 Pro (belief: 3-flash) | 10 | 10,315 | 4,602 | 18,218 | 33,135 | $0.217 |
| `smoke/out-flash37` | 3.7-flash | 10 | 12,636 | 4,660 | 2,946 | 20,242 | $0.038 |


Here's the llm_usage_summary files from those runs:

* [flash-llm_usage_summary.json](https://github.com/user-attachments/files/31533015/flash-llm_usage_summary.json)
* [pro-llm_usage_summary.json](https://github.com/user-attachments/files/31533029/pro-llm_usage_summary.json)

Was hoping to run a larger test on the dev deployment rather than locally.

**Review by dirkraft — APPROVED** [review]

**Comment by dirkraft** [comment]
Heads-up from #68 (litellm routing) — two interactions worth knowing before this merges.

**1. Naming conflict.** #68 requires litellm-qualified model names and rejects unqualified ones at startup, so this PR's `default="gemini-3.7-flash"` would fail:

```
'gemini-3.7-flash' is missing a provider. Model names are litellm-qualified as
<provider>/<model>, e.g. vertex_ai/gemini-3.7-flash or openai/gemini-3.7-flash.
```

Whichever of us merges second needs `vertex_ai/gemini-3.7-flash` (in `args.py`, `easy.py`, `run.py`, `agents.py` and `scripts/run_job.sh`). We also overlap on `standalone.md` and `api/metrics/costs.py`.

**2. Worth measuring before committing to 3.7-flash: its `candidateCount` samples are not independent.** Asking for a random integer in 1–1,000,000 with `temperature=1.0` against Vertex:

| Model | n=5 | n=8 |
| --- | --- | --- |
| `gemini-3.7-flash` | 5 choices, **2 distinct** | 8 choices, **3 distinct** |
| `gemini-3.1-pro-preview` (current default) | 5 → 3 distinct | 8 → 7 distinct |
| `gemini-3-flash-preview` | 5 → 5 distinct | 8 → 8 distinct |

Collisions at that range aren't chance. This matters because `--n_belief_samples` (default 5) becomes `n` on one request, and belief distributions are built from those samples — so 3.7-flash would give noticeably less spread than the model it replaces, on top of the cost win. `gemini-3-flash-preview` is clean, so it's model-specific rather than a Gemini-wide trait.

Caveat on my own measurement: a random-integer prompt is a harsh test and may overstate the effect for real belief prompts, where the output space is constrained. Filed #73 to validate this properly against whatever we settle on as the primary model set — this PR is the main input to that list. Not blocking, just data you'd probably want alongside the pricing case.

**Comment by dirkraft** [comment]
**Retracting my earlier comment on candidate independence — it was wrong.** Apologies for putting unverified reasoning on your PR.

I ran the control I should have run first: one `n=5` request vs five separate `n=1` requests, three trials each.

```
vertex_ai/gemini-3.7-flash    one n=5 call   -> distinct per trial: [4, 5, 4]
vertex_ai/gemini-3.7-flash    5x n=1 calls   -> distinct per trial: [5, 4, 2]
vertex_ai/gemini-3-flash-preview  one n=5 call -> distinct per trial: [5, 4, 3]
```

Two things fall out:

1. **`candidateCount` is not the cause.** Batched and separate requests collide at the same rate — the separate calls actually produced the worst trial (2 distinct). So there is no evidence that asking for 5 candidates in one request gives less independence than 5 requests.
2. **My 3.7-flash vs 3-flash-preview comparison was single-trial noise.** The `5 → 2 distinct` I quoted was one unlucky draw; repeated it gives 4, 5, 4. And `3-flash-preview`, which I called "clean" on the strength of one 5/5 draw, gives 5, 4, 3 on repetition. Both models behave the same.

What I was actually measuring was LLMs being bad at picking random numbers — a strong prior over which digits look random — not a sampling defect. My "collisions at that range aren't chance" line assumed a uniform distribution that was never a given.

**No concern with 3.7-flash from me.** Nothing here should have affected your merge decision, and I'm sorry for the noise. #73 is updated with the retraction.

radamson:
e.g., run from main.autodiscovery.allen.ai using flash (I checked the GCP logs to confirm).

https://autodiscovery.allen.ai/runs/shared/2e8291cf-6544-4114-8bac-809a94bdab35

PR #76, opened 2026-09-02
Cache credits endpoint and skip orphan metadata lookups

### LINKED ISSUE #75: /api/user/me/credits can exceed the 60s proxy timeout — uncached GCS fan-out + orphaned job dirs cause 504s
## Summary

`GET /api/user/me/credits` intermittently exceeds the proxy's 60s upstream timeout, producing 504s on `autodiscovery.allen.ai`. On 2026-09-01 this fired the "10+ 5xx errors in 1 hour" alert: **10× 504, all on `/api/user/me/credits`, all at exactly 60.00s, between 21:40:40Z and 21:42:23Z**, affecting 5 distinct client IPs. nginx logged `upstream timed out (110) while reading response header from upstream` for each. These were the only 5xx on the service in the preceding 7 days apart from one unrelated 500.

## Why it's slow

`get_user_credits()` (`api/utils/credits.py`) does an uncached GCS fan-out on **every** request:

1. `list_user_jobs()` — one GCS list per request.
2. A `ThreadPoolExecutor(max_workers=min(16, len(job_ids)))` that, per job, calls `get_run_details()` **and** `calculate_job_credits()` → `get_job_stats()` — several more GCS round trips each.

Three things compound:

- **The endpoint is polled continuously.** Proxy logs show `/api/user/me/credits` requests at roughly 1/second from many different pages (`/runs`, `/runs/<id>`, `/runs/shared/<id>`). Every open tab re-runs the whole fan-out; nothing is cached or coalesced.
- **The GCS HTTP connection pool is too small for the fan-out.** Sustained `Connection pool is full, discarding connection: storage.googleapis.com. Connection pool size: 10` — 16 worker threads per request, times several concurrent requests, against a 10-connection pool. Discarded connections mean new TLS handshakes, so latency degrades exactly when load rises.
- **Orphaned job directories cost double and always fail.** `list_user_jobs()` returns any `users/<uid>/jobs/<jobid>/` prefix, including ones with no `metadata.json`. `download_job_metadata()` then pays `download_as_text()` → `NotFound` → a *second* GCS call to `job_exists()` → raises `GCSError`, which `_get_job_credit_contribution` swallows with `Failed to get credit contribution for job …`. That work is pure waste and repeats on every poll, forever.

Current volume of that last case in prod: roughly **80–150 `Failed to get credit contribution` warnings per minute, sustained**, spanning **42 distinct job ids across 3 users** in a 10-minute sample. Steady state p50 for the endpoint is ~0.2s, but it has no headroom — when enough concurrent polls land on one instance the worker pool blocks and requests cross the 60s proxy timeout.

## Suggested fixes

Roughly in order of effort/impact:

1. **Cache the result.** A short per-user TTL cache (or a single-flight guard) on `get_user_credits` would collapse N polling tabs into one fan-out and remove most of the load.
2. **Stop re-listing orphans.** Either skip prefixes with no `metadata.json` in `list_user_jobs()`, or have `download_job_metadata()` treat a `NotFound` as a terminal miss without the extra `job_exists()` round trip. Separately, worth checking why those ~42 job directories exist without metadata — partially-created or partially-deleted jobs.
3. **Raise the GCS client's connection-pool size** to at least the thread-pool width (`google-cloud-storage` accepts a custom `requests` session / `urllib3` `pool_maxsize`), so the fan-out stops thrashing connections.
4. **Bound the endpoint's own latency** — a deadline on the fan-out that returns a stale/partial figure rather than letting the request sit until nginx kills it at 60s. A credits readout is not worth a 504.

Reported from a production alert on the prod web stack.

gas2own:
## Summary

- cache `/api/user/me/credits` results for five seconds per user, using stdlib `functools.lru_cache` keyed on a TTL bucket
- keep run-submission credit validation uncached so authorization always uses fresh usage
- skip orphaned job prefixes after a single missing-metadata request instead of issuing a redundant `job_exists()` lookup

## Validation

- `17 passed`: focused credits and GCS tests
- `154 passed, 1 skipped`: full suite excluding `packages/code_execution` (its subprocess tests time out under the dispatcher runtime)
- `ruff check` on the changed files reports only findings that pre-date this branch (`D107`, one `E501` in untouched code)

Closes #75.

Suggested-by: @dirkraft

### DIFF
--- a/api/utils/credits.py
+++ b/api/utils/credits.py
@@ -28,7 +28,6 @@ Example usage:
 
 import logging
 from concurrent.futures import ThreadPoolExecutor
-from dataclasses import astuple
 from functools import lru_cache
 from time import monotonic
 from typing import Any, NamedTuple
@@ -306,15 +305,14 @@ def get_user_credits(userid: str, config: JobConfig | None = None) -> UserCredit
 
 
 @lru_cache(maxsize=CREDITS_CACHE_MAX_ENTRIES)
-def _aggregate_user_credits(
-    userid: str, config_fields: tuple[Any, ...], _ttl_bucket: int
-) -> UserCredits:
+def _aggregate_user_credits(userid: str, config: JobConfig, _ttl_bucket: int) -> UserCredits:
     """Cache key holder for :func:`get_cached_user_credits`.
 
-    ``config_fields`` is a hashable ``astuple`` of the JobConfig; ``_ttl_bucket``
-    changes every ``CREDITS_CACHE_TTL_SECONDS`` so entries age out.
+    ``JobConfig`` is a frozen dataclass, so it is hashable and usable as part of
+    the cache key directly; ``_ttl_bucket`` changes every
+    ``CREDITS_CACHE_TTL_SECONDS`` so entries age out.
     """
-    return get_user_credits(userid=userid, config=JobConfig(*config_fields))
+    return get_user_credits(userid=userid, config=config)
 
 
 def get_cached_user_credits(userid: str, config: JobConfig | None = None) -> UserCredits:
@@ -325,7 +323,7 @@ def get_cached_user_credits(userid: str, config: JobConfig | None = None) -> Use
     """
     config = config or JobConfig()
     ttl_bucket = int(monotonic() // CREDITS_CACHE_TTL_SECONDS)
-    return _aggregate_user_credits(userid, astuple(config), ttl_bucket)
+    return _aggregate_user_credits(userid, config, ttl_bucket)
 
 
 def can_start_experiments(n_experiments: int, userid: str, config: JobConfig | None = None) -> bool:
--- a/api/utils/credits.py
+++ b/api/utils/credits.py
@@ -28,10 +28,10 @@ Example usage:
 
 import logging
 from concurrent.futures import ThreadPoolExecutor
-from threading import Lock
+from dataclasses import astuple
+from functools import lru_cache
 from time import monotonic
 from typing import Any, NamedTuple
-from weakref import WeakValueDictionary
 
 from autodiscovery_jobs import JobConfig
 from autodiscovery_jobs.gcs import (
@@ -55,6 +55,9 @@ DEFAULT_EXPERIMENT_LIMIT = 500
 # deliberately short so the UI remains current while repeated reads share one
 # GCS aggregation. Credit checks before run submission bypass this cache.
 CREDITS_CACHE_TTL_SECONDS = 5.0
+
+# Cap on distinct (user, config) entries kept per worker process.
+CREDITS_CACHE_MAX_ENTRIES = 512
 
 
 class JobStats(NamedTuple):
@@ -87,13 +90,6 @@ class UserCredits(NamedTuple):
     consumed: int
     pending: int
     available: int
-
-
-_credits_cache: dict[tuple[str, str, str | None], tuple[float, UserCredits]] = {}
-_credits_cache_lock = Lock()
-_credits_refresh_locks: WeakValueDictionary[tuple[str, str, str | None], Lock] = (
-    WeakValueDictionary()
-)
 
 
 class InsufficientCreditsError(Exception):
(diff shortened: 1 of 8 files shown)


api/utils/credits.py
```diff
@@ -294,6 +309,38 @@ def _get_job_credit_contribution(job_id: str) -> tuple[int, int] | None:
...
+            if cached is not None and cached[0] > monotonic():
+                return cached[1]
+
+        credits = get_user_credits(userid=userid, config=config)
+        with _credits_cache_lock:
+            _credits_cache[cache_key] = (
+                monotonic() + CREDITS_CACHE_TTL_SECONDS,
+                credits,
+            )
+        return credits
```
dirkraft: @gas2own is there not some library/decorator already available in this project that can do all this so we can avoid some LoC? the answer might be no. do not add any dependencies whatever you find.

gas2own: @dirkraft Answered — and yes, there was a way to cut it. Pushed the simplification.

**What already exists in the project:** no caching library or decorator of its own. The only two precedents are `functools.lru_cache(maxsize=1)` in `packages/autodiscovery_jobs/src/autodiscovery_jobs/auth0.py:38` and the hand-rolled `MetricsCache` in `api/metrics/aggregator.py:541` (stale-while-revalidate, background thread, `fcntl` pod lock, GCS-persisted snapshot) — that one is a single global dataset with a 300s refresh, not reusable for a per-user 5s read. `api/requirements.txt` has no cache dependency, and I added none.

**But stdlib `lru_cache` does cover this**, keyed on a TTL bucket, so I rewrote it in [5bc0027](https://github.com/allenai/asta-autodiscovery/pull/76/commits/5bc0027): the cache is ~20 lines instead of ~36, and the module-level dict, `Lock`, `WeakValueDictionary` of refresh locks, and double-checked read are all gone. It's also now bounded (`maxsize=512`) where the dict grew forever.

Two things that fell out of looking closer:

- **The coalescing was never reachable.** `api/start.sh` runs gunicorn with `--workers 10` and no `--threads` — sync workers handle one request at a time, so a worker can't have two concurrent credits misses for the refresh lock to serialize. That machinery was cost with no effect.
- **Trade-off worth naming:** a bucketed TTL means an entry's effective life is 0–5s rather than exactly 5s (a request landing just before a bucket flip re-aggregates right after). For a polled display that's fine; if you'd rather have an exact per-entry TTL, that's the one thing the hand-rolled version bought.

`17 passed` on the focused credits/GCS tests, `154 passed, 1 skipped` on the full suite excluding `packages/code_execution` (its subprocess tests time out in this runtime). `ruff check api/utils/credits.py` reports the same 4 findings as the base commit — no new ones.

_ran via Claude_

dirkraft: @gas2own not super crazy about the `astuple(config)` and then back again to `JobConfig(*config_fields)`

what would it take to make JobConfig hashable?

gas2own: @dirkraft Not much: `@dataclass(frozen=True)` on `JobConfig` plus one rewrite of `from_env`. Pushed as [178dd32](https://github.com/allenai/asta-autodiscovery/commit/178dd32).

Every field is `str`/`str | None`, and nothing in the repo mutates a `JobConfig` instance — the only assignment was `from_env`'s own override loop, which is now `dataclasses.replace(config, **known)` (still ignoring unknown keys). So freezing gives `__hash__`/`__eq__` for free and the `astuple` → `JobConfig(*config_fields)` round-trip goes away; the cache takes the config object itself:

```python
@lru_cache(maxsize=CREDITS_CACHE_MAX_ENTRIES)
def _aggregate_user_credits(userid: str, config: JobConfig, _ttl_bucket: int) -> UserCredits:
    return get_user_credits(userid=userid, config=config)
```

Two side benefits over the tuple key: field *names* now matter, so reordering or inserting a field can't silently reshuffle values into the wrong parameters the way positional `JobConfig(*fields)` could; and callers get a real immutability guarantee on a config that was already treated as read-only. The cost is that anyone who *did* want to mutate an instance now has to use `replace` — nothing does today.

Added `packages/autodiscovery_jobs/tests/test_config.py` (hashability, immutability, `from_env` overrides) and two credits tests pinning that equal-valued configs share a cache entry and different configs don't. `159 passed, 1 deselected` on the full suite excluding `packages/code_execution`; `ruff check` on the touched files reports the same 4 pre-existing findings as base.

Still open from the last round: the bucketed TTL gives an entry a 0–5s effective life rather than exactly 5s. Say the word and I'll swap in an exact per-entry expiry. I can also fire an independent AI review over the reworked diff if you want a second pass before merge.

_ran via Claude_

packages/autodiscovery_jobs/src/autodiscovery_jobs/config.py
```diff
@@ -3,17 +3,22 @@
...
+from dataclasses import dataclass, replace
 
 # How many days uploaded datasets are retained before the cleanup cron deletes
 # them from GCS.  Both the cleanup script and the API's expiry estimate read
 # this value so they stay in sync.
 DATASET_EXPIRY_DAYS: int = 7
 
 
-@dataclass
+@dataclass(frozen=True)
```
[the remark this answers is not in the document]
dirkraft: true, but i'm ok with this breaking change. nothing in the class semantically invites mutation, and mutation invites strange behavior to modify config after it has been read/constructed. keep frozen

api/utils/credits.py
```diff
@@ -294,6 +304,28 @@ def _get_job_credit_contribution(job_id: str) -> tuple[int, int] | None:
     )
 
 
+@lru_cache(maxsize=CREDITS_CACHE_MAX_ENTRIES)
+def _aggregate_user_credits(userid: str, config: JobConfig, _ttl_bucket: int) -> UserCredits:
```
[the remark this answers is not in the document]
dirkraft: @gas2own debate this point raised by copilot. i don't want more code unless the LoC are worth their weight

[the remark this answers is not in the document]
gas2own: @dirkraft Don't add it. Copilot's mechanism is right in general — `lru_cache` does release its lock around the wrapped call — but the premise that concurrent misses exist here doesn't hold in this deployment.

**Nothing to coalesce in-process.** `api/start.sh` runs `gunicorn --workers 10` with no `--threads` and no `--worker-class`, so it's the default sync worker over a Flask WSGI app: one request per process at a time, no async handler. `api/Dockerfile` pins `ENTRYPOINT ["/api/start.sh"]` and notes Skiff2/Cloud Run can't override it. A worker physically cannot have two credits requests in flight, so a per-key in-flight lock guards a door only one caller can ever walk through.

**The concurrency that does exist is cross-process, and no threading primitive touches it.** Each of the 10 workers holds its own `lru_cache`; the ceiling is 10 aggregations per bucket regardless. That was equally true of the hand-rolled `Lock` + `WeakValueDictionary` version — it's why I called that machinery unreachable when I deleted it in 5bc0027. Copilot is recommending back the exact code that didn't do anything.

**Cross-process coalescing has a known price in this repo.** `api/metrics/aggregator.py`'s `MetricsCache` is what it actually costs: `fcntl` pod lock, background refresh thread, GCS-persisted snapshot, stale-while-revalidate — ~100 LoC and a fresh set of failure modes, for a credits chip. Doesn't pay.

The one thing I'd rather flag than the lock: the residual weakness isn't coalescing, it's **hit rate**. Ten independent caches with a 5s TTL only hit when a user's requests land on the same worker inside the same bucket, so at the ~1 req/s #75 measured this mostly absorbs bursts — page mounts with several consumers, retry pile-ups during a slowdown. That is the 504 scenario, so it's aimed correctly, but don't expect much steady-state relief. If 504s recur, #75's items 3 and 4 (GCS pool size ≥ the 16-thread fan-out; a deadline returning a stale figure instead of sitting until nginx kills it at 60s) buy more per line than anything in-process.

Net: no new code, but I pushed 6 lines of docstring recording the sync-worker reasoning so the next reviewer doesn't re-derive it ([fb07a0c](https://github.com/allenai/asta-autodiscovery/pull/76/commits/fb07a0c)) — happy to drop it for a zero delta. Frozen `JobConfig` stays per your call.

Still open: the bucketed TTL gives an entry 0–5s of effective life rather than exactly 5s — say the word and I'll swap in exact per-entry expiry.

_ran via Claude_

dirkraft (approved):

PR #80, opened 2026-09-03
Add maintainer-only per-user data purge for GDPR erasure

### LINKED ISSUE #72: AutoDiscovery/DataVoyager: add per-user data purge for GDPR erasure
## AutoDiscovery / DataVoyager: add a per-user data purge (GDPR erasure)

To support GDPR "right to be forgotten" deletion, we need a way to erase all of a single user's AutoDiscovery/DataVoyager data. Today deletion is only job-scoped, so a user's data survives across jobs.

**Where a user's data lives:**
- Primary GCS `autodiscovery` bucket: `users/{sub}/jobs/{jobid}/...` (uploaded datasets, results, credits), keyed by Auth0 `sub` (`gcs.py:74,531`, `api/runs/runs_api.py:963,1555`).
- Copies in the Asta workspaces bucket under `owners/{uuid}/{tid}/data/` (`manager.py:257-313`, `asta_gcs.py:51`).

**Today:** deletion is job-scoped only — `soft_delete_job`, and a `delete_job_directory` helper that exists but is **not routed** to any user-facing path — plus a 7-day dataset TTL. There is no user-level purge.

**Ask:** wire a purge-by-`sub` that removes `users/{sub}/` in the primary bucket and the Asta-copy `owners/{uuid}/` prefix for a subject, so a right-to-be-forgotten deletion covers uploaded datasets, results, and credits.

Part of a cross-repo Asta/Nora GDPR "right to be forgotten" effort.

Suggested-by: @charliemcgrady

gas2own:
Closes #72.

Deletion in AutoDiscovery was job-scoped only — `soft_delete_job` (which keeps results and metadata) plus a 7-day dataset TTL — so a user's data survived across jobs and there was no way to satisfy a GDPR right-to-be-forgotten request.

Per the design in #72, this adds the erasure path at the data-access layer plus one interactive maintainer script that calls it. **There is no HTTP route and no `JobManager` method** — the purge preserves nothing and cannot be undone, so the only intended caller is a maintainer at a terminal.

Scope is AutoDiscovery's own bucket. Data AutoDiscovery hands to another system is erased by that system, not from here.

### `autodiscovery_jobs.gcs`

- `summarize_user_data(sub)` inventories a subject's footprint without touching it: object count and bytes, job IDs, credits profile, shared-run index entries, and any job still in a non-terminal state.
- `purge_user_data(sub, dry_run=...)` deletes all of `users/{sub}/**` plus the user's `index/shared-runs/` entries.
- Both reject a `userid` containing `/`, so a malformed subject cannot widen the delete prefix beyond one user.

The shared-run index is scanned in full rather than derived from the user's job directories: an entry can outlive its job directory, and the entry body itself stores the `userid`.

### `scripts/purge_user_data.py`

```bash
# Inventory only — deletes nothing, prompts for nothing
uv run python scripts/purge_user_data.py --sub 'google-oauth2|123' --dry-run

# Purge — prints the same inventory, then requires the sub to be retyped
uv run python scripts/purge_user_data.py --sub 'google-oauth2|123'
```

Three guards before anything is deleted: the confirmation requires the exact `sub` retyped on a TTY (it refuses rather than proceeding unattended without one); it refuses while any of the subject's jobs is in a non-terminal state, since a still-running job would write new data after the purge; and `--dry-run` gives the full inventory with no prompt.

It also prints, on every run, the surfaces it *cannot* reach so they are not assumed covered — the dataset copies handed to the Asta workspaces bucket under `owners/{uuid}/` and the asta-context-service artifacts registered from them, the Asta user record and threads, the Auth0 profile, and application logs — each of which needs its own erasure request in the cross-repo effort. Separately it notes the one surface that needs no action: the metrics dashboard's job snapshot is derived by rescanning the job directories, so the subject's rows fall out of it on the next refresh.

### Verification

`make test` passes (92 in `autodiscovery_jobs`, 198 repo-wide). 11 tests in `packages/autodiscovery_jobs/tests/test_user_purge.py` cover the inventory across every surface, full and dry-run purges, active-job detection, prefix-widening rejection, and error wrapping. One pre-existing failure in `packages/code_execution/tests/test_ipython_session.py::test_run_cell_subprocess_success` also fails on a clean checkout of `main` and is untouched by this change.

Co-authored-by: @dirkraft


packages/autodiscovery_jobs/src/autodiscovery_jobs/asta_gcs.py
```diff
@@ -61,4 +61,120 @@ def copy_dataset_to_asta_workspace(
...
+            blob.delete()
+            _log.info("Purged gs://%s/%s", ASTA_BUCKET, blob.name)
+
+    return {
+        "user_uuid": user_uuid,
+        "bucket": ASTA_BUCKET,
+        "dry_run": dry_run,
+        "deleted_objects": sorted(deleted_objects),
+        "deleted_bytes": deleted_bytes,
+    }
```
dirkraft: omit these changes to delete from the workspace bucket. This isn't super clear as designed now, but the idea is that we are clients to another system which has this workspace bucket. Once the data has been copied and the call has been made to start a session on the workspace data in another system, we (AutoDiscovery) are no longer involved.

packages/autodiscovery_jobs/src/autodiscovery_jobs/gcs.py
```diff
@@ -1131,3 +1132,310 @@ def generate_upload_url(
...
+# "right to be forgotten" request. It is deliberately NOT surfaced on JobManager
+# or any HTTP route: the only intended caller is scripts/purge_user_data.py,
+# run by a system maintainer. See scripts/README.md.
+# ---------------------------------------------------------------------------
+
+# Derived metrics snapshot. It caches one record per job -- including the owning
+# userid -- so a subject erasure has to strip the subject's rows from it too.
+# api/metrics/aggregator.py imports this as its PERSIST_BLOB_PATH so the path has
+# a single definition.
+METRICS_CACHE_BLOB_PATH = "_metrics/jobs_cache.json"
```
dirkraft: I'm wondering what's the consequence of doing no deletion maintenance to this metrics data? If it's not absolutely necessary I'd rather not mess with it.

dirkraft:
@gas2own please address the two comments above

gas2own:
@dirkraft Both addressed and pushed as [a8580c2](https://github.com/allenai/asta-autodiscovery/pull/80/commits/a8580c2) — the branch now touches only `gcs.py`, its tests, the script, and `scripts/README.md`.

**Workspace bucket** — removed entirely. `summarize_asta_workspace_data`/`purge_asta_workspace_data` and the `--asta-user-uuid` flag are gone, and `asta_gcs.py` is back to its state on `main`. The script still *names* those copies in its closing notes, but under "close these out separately — owned by Asta", not as something AutoDiscovery deletes.

**Metrics cache** — also removed, and your instinct was right that it wasn't necessary. That snapshot is fully derived: `_scan_all_jobs` rediscovers jobs with a glob over `users/*/jobs/*/run_details.json` and carries a cached row forward *only* if the job is rediscovered, then overwrites the persisted blob with the rebuilt set. So once the job directories are purged, the subject's rows drop out on the next refresh (5-min staleness window, on dashboard traffic) and cannot come back. Editing the blob only shortened that window, and it made the purge script a second writer on a blob the API pods own. `aggregator.py` keeps its own `PERSIST_BLOB_PATH` — no shared constant, no import from `gcs`. The script prints the residual explicitly under a "no action needed" heading so the next maintainer doesn't wonder.

Tests: 92 in `autodiscovery_jobs`, 198 repo-wide, same one pre-existing `test_ipython_session` failure that fails on clean `main`. PR body updated to the narrower scope.

_ran via Claude_

dirkraft (approved):

PR #81, opened 2026-09-03
Always interpret plots outside the code-execution environment

### LINKED ISSUE #74: Always interpret plots outside the code-execution environment
Raised by @radamson while reviewing #68. Related: #55 (unify code execution on the asta-sandbox backend registry), which this should probably ride along with rather than precede.

## Problem

Plot interpretation happens in two completely different places depending on `--backend`, and one of them runs the vision model **inside the code-execution environment**.

### Current state

**`--backend process` / `modal` — analysis outside the sandbox (good)**

The sandbox returns figures as structured rich outputs. `ModalSandboxExecutor.execute_code_blocks` picks base64 PNGs out of `result.rich_outputs` and calls the vision model from the parent process (`_analyze_image`). Figures are then persisted by `run._write_rich_outputs` to `rich_outputs/ro_<level>_<idx>.json` and rendered into the HTML report.

**`--backend local` — analysis inside the executed code (bad)**

`CodeBlockWrapperTransform` prepends a generated `build_image_analysis_patch(...)` block to every code block the agent runs. That patch monkeypatches `plt.show` so that, inside the executed program:

- it re-encodes each open figure to base64,
- calls the vision model **from within the execution environment**,
- prints the analysis to stdout,
- and emits a token-usage record as a magic stdout line (`__AUTODISCOVERY_LLM_USAGE__` + JSON), which the parent later scrapes back out via `llm_usage.extract_local_image_usage_markers`.

Consequences specific to this path:

| | |
|---|---|
| Credentials | model credentials must be reachable *from inside* the execution environment |
| Trigger | only fires on `plt.show()`; a figure that is saved or left open is never analysed |
| Transport | analysis and usage travel as stdout text, with usage parsed from a marker string |
| Persistence | figures are **not** stored — `_get_executor_rich_outputs` returns `[]` because `LocalCommandLineCodeExecutor` has no `get_last_rich_outputs`, so no `rich_outputs/` entries and nothing in the report |
| Injection | agent-authored code is rewritten before execution, so what runs is not what the agent wrote |

So `local` is both the least isolated backend *and* the one that needs credentials inside the sandbox — the inverse of what you'd want.

## Desired state

**One path: the execution environment returns figures as data; the parent interprets them.**

- Every backend surfaces figures through the same rich-output channel, so `_analyze_image` in the parent is the only place a vision model is called.
- `build_image_analysis_patch`, `CodeBlockWrapperTransform`, `LOCAL_IMAGE_USAGE_MARKER` and `extract_local_image_usage_markers` all disappear. No code injection, no stdout scraping, no credentials in the sandbox.
- Figures are persisted for every backend, so the report and `rich_outputs/` are backend-independent.
- Because analysis is decoupled from `plt.show`, it can be more comprehensive than today: analyse every figure produced, and re-analyse from stored plots without re-running the experiment. That last part is @radamson's point — the plots are already on disk for the sandbox backends, so interpretation no longer has to happen inline with execution.

## Why this is cheap now

#68 already removed the worst of the duplication: the injected patch used to carry its own copy of Vertex auth, token refresh and model-name normalisation, and now just calls `autodiscovery.llm`. So the patch is small, and deleting it is mostly a matter of `local` gaining rich outputs.

## Sequencing

This is really a corollary of #55. That issue notes `local` is autogen's `LocalCommandLineCodeExecutor` while `process` is the workspace's own `ProcessIPythonBackend` — and it is precisely the IPython-kernel-shaped backends that produce rich outputs. Once `local` routes through the asta-sandbox registry (`inprocess`), it gets rich outputs for free and the patch has no reason to exist.

**Worth asking as part of #55: should `--backend local` exist at all?** `process` is also local — a subprocess on the same machine — just isolated, and it is the default. `local` differs only by having no isolation, and it is already unsupported for Copilot (`get_agents` rejects that combination). If `local` were dropped, this issue closes by deletion rather than by implementation.

## Acceptance

- [ ] Vision model is never invoked from inside the execution environment, on any backend.
- [ ] `build_image_analysis_patch` / `CodeBlockWrapperTransform` deleted, or `local` removed.
- [ ] `LOCAL_IMAGE_USAGE_MARKER` and its stdout-scraping parser deleted; image-analysis usage is recorded through `UsageTracker` like every other call.
- [ ] Figures persist to `rich_outputs/` and appear in the report regardless of backend.
- [ ] Analysis no longer depends on the agent calling `plt.show()`.

Also worth folding in @radamson's broader note that "the way figures are handled by autodiscovery in general isn't great" — the above is the structural half of that; figure quality and selection is a separate conversation.

gas2own:
Closes #74.

## What changed

Plot interpretation happened in two different places depending on `--backend`, and one of them ran the vision model **inside** the code-execution environment. `--backend local` had `build_image_analysis_patch` prepended to every code block: it monkeypatched `plt.show` so the executed program itself base64-encoded each figure, called the vision model, printed the analysis to stdout, and emitted token usage as a magic marker line the parent scraped back out.

Now there is one path. `local` routes through `code_execution.LocalIPythonBackend`, adapted by the same wrapper `process` already used, so every backend returns figures as rich outputs and `SandboxCodeExecutor._analyze_image` is the only place a vision model is called.

Deleted: `build_image_analysis_patch`, `CodeBlockWrapperTransform`, `LOCAL_IMAGE_USAGE_MARKER`, `extract_local_image_usage_markers`, and the usage-scraping block in `run.py`. Net −140 lines of source.

`ModalSandboxExecutor` → `SandboxCodeExecutor`, since all three backends now use it.

## Acceptance

| | |
|---|---|
| Vision model never invoked inside the execution environment | ✅ only `SandboxCodeExecutor._analyze_image`, in the parent |
| `build_image_analysis_patch` / `CodeBlockWrapperTransform` deleted | ✅ |
| `LOCAL_IMAGE_USAGE_MARKER` + parser deleted; usage via `UsageTracker` | ✅ one `image_analysis` component; `image_analysis.local` / `.modal` are gone |
| Figures persist to `rich_outputs/` on every backend | ✅ `local` gained `get_last_rich_outputs`, which is what `run._get_executor_rich_outputs` keys off |
| Analysis no longer depends on `plt.show()` | ⚠️ done for `local` + `process`; see below |

`IPythonSession` now publishes figures a cell left open once the cell finishes, so a figure that was built and saved — or simply forgotten — is still returned, and a shown figure is not published twice. That covers `local` and `process`. **`modal` still only publishes on `plt.show()`**: its kernel is asta-sandbox's, not this repo's. Worth a follow-up against asta-sandbox, and #55 is the natural place for it since it moves `process` into that library too.

## On `--backend local`

The issue asks whether `local` should exist at all, which would close this by deletion instead. I kept it, because that call belongs with #55 — and this change is worth having either way, since it also gives `local` persisted figures and lifts the `github_copilot` restriction. If `local` does get dropped, this PR makes that a pure deletion rather than an untangling.

`local` keeps its character: no separate sandbox venv, so generated code sees the job's own packages. It still runs each cell in a child process — which is what AG2's `LocalCommandLineCodeExecutor` gave it and what makes the timeout enforceable — so it is not more exposed than before.

## Validation

`make test` — 193 passed. `test_run_cell_subprocess_success` fails in my container and fails identically on `main`: its 2s timeout is not enough for a `spawn` here. Not touched by this change; the same subprocess path works with a longer timeout. Worth confirming it is green in CI.

New tests cover: `local` returns rich outputs; nothing is injected into the executed code; the deleted symbols stay deleted; the adapter marshals rich outputs and honors `use_subprocess`; every returned PNG is analyzed once in the parent with usage recorded; and a figure left open is captured while a shown one is not double-published.

`ruff format` reformats 23 unrelated files on this repo, so I reverted everything outside this change. `pyright` reports one error in `agents.py` that is also on `main`.

Suggested-by: @radamson
Suggested-by: @dirkraft


dirkraft:
@gas2own resolve conflicts with main

gas2own:
@dirkraft Conflicts resolved and pushed — [PR #81](https://github.com/allenai/asta-autodiscovery/pull/81) is `MERGEABLE` and all CI checks are green (including the `test_run_cell_subprocess_success` that fails only in my container). You're still the requested reviewer; nothing else is open on this PR.

_ran via Claude · shortened — ask for the full reply_

packages/autodiscovery/src/autodiscovery/agents.py
```diff
@@ -859,19 +776,28 @@ def install(package):
...
+            _IPythonBackendAdapter(process_backend),
             timeout=code_timeout,
             vision_model=vision_model,
             usage_tracker=usage_tracker,
         )
         print(f"Using process backend with work_dir: {work_dir}")
     else:
-        # Use local code executor (in-process, no isolation)
-        executor = LocalCommandLineCodeExecutor(
+        # Local: the job's own Python environment, with no per-cell venv.
```
dirkraft, line 787: assert this if you're not going to check it explicitly

dirkraft (changes requested):
@gas2own does this mean that code execution environments do not need (vision) model credentials plumbed? it would be nice if we could say that code execution environments (besides same host/process) do not by default inherit any llm provider credentials. is there any plumbing code that existed solely to plumb credentials from AD jobs across a boudnary to code execution environments? if so, consider if it can be retired/removed. it's possible there is no such code to retire.

gas2own:
Started async job `gas2own-d9d26ab9e6b5` (1/10 slots in use). I’ll post its result here when it finishes.

gas2own:
@dirkraft **Yes — no code-execution environment needs model credentials after this PR.** The only vision call is `SandboxCodeExecutor._analyze_image`, in the parent. And the invariant you want is now enforced, not just true by accident: the `process` backend used to copy the whole job environment into each cell's child process, so an `OPENAI_API_KEY` it had no use for came along. Pushed [519febd](https://github.com/allenai/asta-autodiscovery/commit/519febd): cells start from the job environment minus known provider credentials, with `env=` applied after the scrub as the escape hatch. `modal` never received any (it only gets `DATASET_ROOT` plus the bucket-read secret); `local` is the same-host/process case you excluded, untouched. Also added the assert you asked for at `agents.py:787`.

On plumbing to retire: there was none as such — because the old design avoided the boundary rather than crossing it. `build_image_analysis_patch` ran the vision call *inside* the job's own process specifically so it could `import autodiscovery.llm` and reuse the parent's provider routing and auth; its docstring said as much. That patch is the thing being retired, and this PR already deletes it.

Two things to flag: **unit tests were red on this PR** — the merge with main brought `test_code_output_truncation.py`, which still imports `ModalSandboxExecutor` under its pre-rename name; that's fixed in this push, so my earlier "all checks green" was wrong once that merge landed. And an independent review is running on the PR.

**Decisions needed**
- `GOOGLE_APPLICATION_CREDENTIALS` is in the scrub list because it's how `vertex_ai` authenticates, but it's dual-use — after this, a cell wanting GCP access for anything other than a model must get it passed explicitly. Say the word and I'll drop it from the list.

_ran via Claude · shortened — ask for the full reply_

gas2own:

## P0 (blocking)

- `packages/autodiscovery/src/autodiscovery/agents.py:143` always passes `use_subprocess` to `ProcessIPythonBackend.run_cell`, whose interface does not accept that argument; the default `process` backend will fail with `TypeError`. Only include the keyword for backends supporting it, or normalize backend interfaces.

## P1 (should fix)

- `packages/autodiscovery/src/autodiscovery/agents.py:789` runs local subprocesses with the full parent environment, exposing provider credentials to generated code despite credentials no longer being required there. Scrub credentials for the local backend too.
- `packages/code_execution/src/code_execution/ipython_session.py:92` republishes every open figure, including figures already emitted through `display(fig)` or as a cell’s final expression, causing duplicate outputs, analyses, and charges. Track emitted figure identities and flush only unseen figures.

## P2 (nits)

- `packages/autodiscovery/src/autodiscovery/agents.py:786` validates an unknown backend with `assert`, which disappears under `python -O`. Raise `ValueError` explicitly.

gas2own:
Addressed the independent review in [67ebb1f](https://github.com/allenai/asta-autodiscovery/commit/67ebb1f):

- **`use_subprocess` / process backend:** no change; this is a false positive. `ProcessIPythonBackend.run_cell` currently accepts `use_subprocess`, and the adapter test exercises that call.
- **Local backend credential inheritance:** no change; `local` is deliberately the same-host/process exception to the credential-isolation requirement. Whether to tighten that exception remains a policy decision rather than a correctness fix for this PR.
- **Duplicate figures after `display(fig)`:** fixed. The session now tracks objects already formatted during the cell and flushes only unseen open figures; a regression test covers explicit display.
- **Unknown backend assertion:** fixed. Unknown values now raise `ValueError` even under optimized Python.

Validation: 228 tests passed. The one failure is the already documented `test_run_cell_subprocess_success` 2-second startup timeout that also fails on `main` in this environment. Focused affected tests pass (21/21), and pyright reports only the pre-existing `agents.py:402` error.

gas2own:
Started async job `gas2own-7942fa33b5bc` (2/10 slots in use). I’ll post its result here when it finishes.

gas2own:

## P0 (blocking)

*   `packages/autodiscovery/src/autodiscovery/agents.py:144`: `_IPythonBackendAdapter.run_code` unconditionally passes `use_subprocess` to `self._backend.run_cell()`. However, `ProcessIPythonBackend.run_cell` does not accept this keyword argument, which will cause a `TypeError` and crash the default `process` backend on every run. (Note that the mock in `test_agents_backends.py` was updated to accept it, masking the bug).
    *   **Fix:** Remove `use_subprocess` from `_IPythonBackendAdapter`'s signature entirely and configure it on `LocalIPythonBackend` during its instantiation. Alternatively, build a kwargs dict and conditionally pass it: `kwargs = {"timeout_s": timeout_seconds}; if self._use_subprocess: kwargs["use_subprocess"] = True; result = self._backend.run_cell(code, **kwargs)`.

## P1 (should fix)

*   `packages/code_execution/src/code_execution/ipython_session.py:90-100`: The `try...except Exception` block in `_flush_open_figures` wraps the entire `for` loop. If `display(figure)` fails on a figure, the function returns early, failing to display or close any subsequent open figures, leading to memory leaks.
    *   **Fix:** Move the exception handling and closure inside the loop so each figure is processed independently:
        ```python
        for fig_num in pyplot.get_fignums():
            figure = pyplot.figure(fig_num)
            try:
                if id(figure) not in formatted_object_ids:
                    display(figure)
            except Exception:
                pass
            finally:
                pyplot.close(figure)
        ```

## P2 (nits)

*   `packages/code_execution/src/code_execution/process_backend.py:83`: `_PROVIDER_CREDENTIAL_ENV_VARS` misses `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. Since Bedrock is a common LLM provider, these credentials should also be scrubbed from the sandbox environment unless explicitly requested.
    *   **Fix:** Add AWS credentials to the `frozenset`.

gas2own:
TL;DR: The [review was blocking](https://github.com/allenai/asta-autodiscovery/pull/81#issuecomment-5610416427). I pushed [67ebb1f](https://github.com/allenai/asta-autodiscovery/commit/67ebb1f), posted the [finding-by-finding resolution](https://github.com/allenai/asta-autodiscovery/pull/81#issuecomment-5610457502), and all CI is green. The alleged process-backend P0 was a false positive; duplicate figure analysis and optimized-away validation were fixed. A rotated re-review is running.

**Decisions needed**

- Keep `local` as the same-host credential-inheritance exception, or scrub provider credentials there too?
- Keep `GOOGLE_APPLICATION_CREDENTIALS` scrubbed from process backends, or exempt it because it also enables non-model GCP access?

No knowledge-base update was warranted.

_ran via Codex_

gas2own:
Addressed the [Gemini review](https://github.com/allenai/asta-autodiscovery/pull/81#issuecomment-5610485312) in [79fa693](https://github.com/allenai/asta-autodiscovery/commit/79fa693):

- **`use_subprocess` / process backend:** no change; false positive. The concrete `ProcessIPythonBackend.run_cell` signature accepts `use_subprocess`, its backend tests execute that method, and the adapter test passes the keyword through.
- **A failed figure render stops the flush:** fixed. Exception handling and closure are now per figure, so one failed display cannot prevent subsequent figures from being published or closed. A regression test verifies both later progress and cleanup.
- **AWS credentials:** no change. `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are general-purpose AWS credentials, not Bedrock-specific, so adding them would also remove ordinary data-access authority from generated analysis code. That is a broader sandbox policy choice, like the already-open `GOOGLE_APPLICATION_CREDENTIALS` question, rather than a correctness fix to apply implicitly here.

Validation: focused figure-flush and backend-adapter tests pass; Ruff check and format check pass. The full non-Modal/non-ADC suite has 229 passes and the one already documented `test_run_cell_subprocess_success` two-second startup-timeout failure that also occurs on `main` in this environment. No independent re-review was fired for this small follow-up diff.

gas2own:
TL;DR: The [blocking re-review](https://github.com/allenai/asta-autodiscovery/pull/81#issuecomment-5610485312) found one valid bug, fixed in [79fa693](https://github.com/allenai/asta-autodiscovery/commit/79fa693). The repeated backend P0 was false; AWS credential scrubbing remains a policy choice. [Resolution posted](https://github.com/allenai/asta-autodiscovery/pull/81#issuecomment-5610531484), all CI is green, and human re-review is requested. No knowledge update was warranted.

**Decisions needed**

- Should same-host `local` execution inherit provider credentials?
- Should general-purpose GCP/AWS credentials be scrubbed despite their non-model data-access uses?

_ran via Codex_

packages/code_execution/src/code_execution/process_backend.py
```diff
@@ -197,7 +241,9 @@ def run_cell(
             "matplotlib_backend": matplotlib_backend,
         }
 
-        child_env = os.environ.copy()
+        # Inherit the job's environment minus LLM-provider credentials, then
+        # apply any explicitly configured vars (which may re-add one).
+        child_env = strip_provider_credentials(os.environ.copy())
```
dirkraft: @gas2own i'm not excited about (maintaining/validating) this particular list of env vars being stripped. I think we shouldn't attempt to create any special isolation of env vars here, when the rest of the host is still available; the measure is ineffective at best (can access the machine anyways), and misleading or broken at worst (stripped the wrong env var, other machine-local authentication mechanisms still work, ...)

gas2own: @dirkraft I'm still working on this — the inline turn hit its time limit, so I moved it to the long runtime. I'll post the result here when it finishes.

_ran via Claude_

gas2own: @dirkraft Agreed — reverted the scrub in [8d65dac](https://github.com/allenai/asta-autodiscovery/commit/8d65dac). Cells inherit the job environment again; `_PROVIDER_CREDENTIAL_ENV_VARS`, `strip_provider_credentials`, and the two tests asserting the scrub are gone. All CI green on #81, and it's `MERGEABLE`.

Your reasoning holds: the process backend runs on the same host as the same user, so code that wanted a key could read the parent's `/proc/<pid>/environ`, `~/.aws/credentials`, the gcloud ADC file, or instance metadata regardless. A name list buys no boundary, has to be kept correct forever, and invites reading the backend as isolated when it isn't. The property that motivated it stands on its own without the list: nothing inside any execution environment calls a model, because figure interpretation runs in the parent.

This also closes the two questions I'd left open — `GOOGLE_APPLICATION_CREDENTIALS` and the AWS keys — and the `local`-vs-`process` inconsistency: there's now one behaviour across backends. No independent re-review fired; the diff is a deletion within already-reviewed code.

Still open from earlier in the thread: `modal` publishes figures only on `plt.show()`, since its kernel is asta-sandbox's rather than this repo's — I'd take that to asta-sandbox alongside #55, which moves `process` into that library too.

_ran via Claude · shortened — ask for the full reply_

dirkraft: @gas2own leave a quick note on #55 if it isn't there already so that when an agent picks up that task it considers this point about plt.show bheavior specific to modal code executor

gas2own: @dirkraft Added the [note to #55](https://github.com/allenai/asta-autodiscovery/issues/55#issuecomment-5668875978), covering Modal’s `plt.show()` dependency and parity requirements. No knowledge-base update was warranted.

_ran via Codex_

dirkraft:
todo: validate myself

    auto-discovery --name "test" --description "seattle pet licenses" --n_experiments 4 --out_dir ./results/ ~/Downloads/Seattle_Pet_Licenses_20260126.csv --backend local

dirkraft (approved):

Issue #79, opened 2026-09-03
run_mcts spins forever when the data-loader node can't be created, writing GBs of log


dirkraft:
Found while smoke-testing the 1.0.0 release from PyPI. **Not a 1.0.0 regression** — the code is identical at 0.2.2 (`3e976267`). Any durable model failure at the data-loader step triggers it; #78 just made it easy to reach.

## Symptom

A run whose model calls fail persistently does not exit. It emits, at full speed with no backoff:

```
SAMPLED 1 NODE(S) FOR EXPANSION: ['0_0']
==========================
(1/1): EXPANDING NODE 0_0
No new experiment generated for node 0_0. Skipping this iteration.
```

Measured: **3.9 GB of log in 90 seconds** (~204M lines), of which exactly 3 lines were the underlying error. The real failure is unfindable, and an unattended run will fill the disk.

## Mechanism

[`mcts_utils.py:271`](../blob/v1.0.0/packages/autodiscovery/src/autodiscovery/mcts_utils.py#L271):

```python
# If the data loader node hasn't been executed, return the root. This is not run in batch mode.
if len(nodes_by_level[1]) == 0:
    return [root]
```

The root's only job is to produce the data-loader node at level 1. If the model call that generates it fails, `nodes_by_level[1]` stays empty, so `select_nodes` returns `[root]` **unconditionally and forever**. This early return never consults [`has_untried_experiments()`](../blob/v1.0.0/packages/autodiscovery/src/autodiscovery/mcts.py#L194), so the `allow_generate_experiments = False` guard set at [`run.py:541`](../blob/v1.0.0/packages/autodiscovery/src/autodiscovery/run.py#L541) has no effect on this path.

Meanwhile the outer loop at [`run.py:490`](../blob/v1.0.0/packages/autodiscovery/src/autodiscovery/run.py#L490):

```python
while n_sampled < total_to_sample:
```

never advances, because the node is never committed — the failure path `return None`s at `run.py:546` without incrementing `n_sampled`. The `if not next_nodes: break` escape at `run.py:503` is unreachable, since `select_nodes` always returns a non-empty list. No iteration cap, no consecutive-failure counter, no sleep.

## Triggers

Anything that makes data-loader generation fail durably rather than transiently:

- misconfigured or missing Vertex project/location (#78)
- expired or insufficient credentials
- quota exhaustion, or a model the project has no access to
- a model whose responses never parse into an experiment

## Fix

[...]

Issue #78, opened 2026-09-03
Vertex AI: VERTEX_LOCATION no longer defaults to `global`, and a missing project fails silently


dirkraft:
Found while smoke-testing the 1.0.0 release from PyPI. Both are regressions introduced by #68, and the second one means the **published** docs describe behavior the package doesn't have.

## Problem

[`llm.py:339`](../blob/v1.0.0/packages/autodiscovery/src/autodiscovery/llm.py#L339) passes Vertex settings to litellm only when the env var happens to be set:

```python
def _provider_kwargs(model: str) -> dict[str, Any]:
    if provider_of(model) != VERTEX_AI:
        return {}
    kwargs: dict[str, Any] = {}
    if project := os.getenv(VERTEX_PROJECT_ENV_VAR):
        kwargs["vertex_project"] = project
    if location := os.getenv(VERTEX_LOCATION_ENV_VAR):
        kwargs["vertex_location"] = location
    return kwargs
```

### 1. `VERTEX_LOCATION` unset no longer means `global`

0.2.2's `vertex_config.get_vertex_openai_base_url()` did:

```python
location = os.getenv(VERTEX_LOCATION_ENV_VAR) or "global"
```

Now, unset means the key is simply absent and litellm applies its own default of `us-central1` — where the default model `vertex_ai/gemini-3.7-flash` is not served. The result is a 404 that reads like the model doesn't exist:

```
litellm.NotFoundError: Vertex_aiException - {
  "error": {
    "code": 404,
    "message": "Publisher model `projects/<adc-quota-project>/locations/us-central1/publishers/google/models/gemini-3.7-flash` was not found or your project does not have access to it."
```

Both `packages/autodiscovery/RELEASE.md` (the PyPI long description) and `docs/autodiscovery/standalone.md` still say `VERTEX_LOCATION` is *"optional; defaults to global"*. As published, that is not true.

### 2. `VERTEX_PROJECT_ID` unset fails silently

0.2.2 raised immediately and named the variable:

> Vertex AI configuration is required for Gemini models. Set `VERTEX_OPENAI_BASE_URL` or both a project ID (`VERTEX_PROJECT_ID`) and an optional location (`VERTEX_LOCATION`, defaults to global).

Now nothing validates it. litellm falls back to whatever project Application Default Credentials happens to carry, so the 404 above names a project the operator never selected. Anyone whose ADC quota project differs from their Vertex project gets a confusing error pointing at the wrong place.

[...]

PR #83, opened 2026-09-03
Configure Vertex with litellm's own env vars and require both at startup

gas2own:
Fixes both halves of #78 — the two 1.0.0 Vertex regressions from #68, which land on the path a first run takes since the default models are Vertex.

Revised per review: rather than fixing this package's mapping of its own `VERTEX_PROJECT_ID` / `VERTEX_LOCATION` onto litellm's parameters, the mapping is gone. litellm already reads these two settings from the environment under its own names, so the package now checks those names directly and does nothing else with them.

## What was wrong

`_provider_kwargs` passed each Vertex setting to litellm only when its env var happened to be set:

- **Unset `VERTEX_LOCATION` stopped meaning `global`.** The key was simply absent, so litellm applied its own default of `us-central1` (`vertex_llm_base.get_vertex_region` returns `vertex_region or "us-central1"` for a model litellm has not mapped). `vertex_ai/gemini-3.7-flash` is not served there, so the call came back as a 404 reading as if the model did not exist.
- **Unset `VERTEX_PROJECT_ID` stopped being an error.** Nothing validated it, so litellm fell back to the Application Default Credentials project and the 404 named a project the operator never chose.

Both surfaced as durable *per-call* errors mid-run rather than at startup, which is what makes them compound with #79.

## What this does

**Breaking: `VERTEX_PROJECT_ID` and `VERTEX_LOCATION` are no longer read. Set `VERTEXAI_PROJECT` and `VERTEXAI_LOCATION`, both of them.**

litellm resolves both itself, so nothing is forwarded, defaulted or inferred here — `_provider_kwargs`, `_vertex_project`, `_vertex_location` and `VERTEX_DEFAULT_LOCATION` are all gone, and with them the location default and the precedence order between two sets of names for one setting. The package checks the two variables directly in `llm.validate()`.

That startup check is what remains, and it is the point: each litellm fallback is a silent mid-run 404 rather than an error. `llm.validate()` — where every other model precondition already lives, reached from `run.resolve_model_args` — raises `ModelError` before the first model call when either variable is absent. The rename is therefore safe to make loudly: a deployment still setting only the old names fails at startup naming the new ones instead of drifting onto a fallback.

`easy.cli_main` reports a `ModelError` as a one-line flag error and checks the model flags *before* creating a working directory, which also covers the `<provider>/<model>` prefix error every 0.2.x command line hits on upgrade.

### Why `VERTEXAI_` and not litellm's shorter aliases

litellm accepts `VERTEX_PROJECT` / `VERTEX_LOCATION` too, so "which of litellm's names?" is a fair question of a check that rejects one set. It reads the short pair only in `embedding()`; `_complete_vertex_ai` and `_complete_vertex_ai_beta` read the `VERTEXAI_` pair alone (`litellm/main.py`, 1.97). The `VERTEXAI_` pair is the only configuration every call this package makes can see, so accepting the aliases would let a run start on settings its completions cannot read. Covered by `test_litellms_short_aliases_do_not_satisfy_the_check`.

Renames also land in the Cloud Run deploy script, the docker jobs env passthrough, and `.env.example` (the template `docker compose` users copy to `.env`, so leaving the old names there would have produced the new startup error on a fresh checkout). The deploy script's location default moves from `$REGION` (`us-west1`) to `global` alongside — `us-west1` does not serve the default models either, so it was the same bug pinned to a different region.

## Acceptance

- [x] Vertex configuration uses litellm's own variables, mapped and defaulted nowhere.
- [x] A `vertex_ai/` model missing either variable fails at startup naming both required settings — `test_a_vertex_model_needs_both_settings` (parametrized over each) and `test_a_missing_vertex_setting_stops_a_run_before_any_model_call`.
- [x] A run naming no Vertex model needs no Vertex settings — `test_a_run_naming_no_vertex_model_needs_no_vertex_settings`.
- [x] `ModelError` is caught in `easy.cli_main` and exits with the message — `test_easy_cli_reports_a_model_error_as_a_flag_error`, parametrized over the missing-project and missing-prefix cases.
- [x] `RELEASE.md`, `standalone.md`, `docs/configuration.md`, `.env.example` and a `CHANGELOG.md` entry match actual behavior.

## Verification

The issue's exact repro, with every Vertex env var unset and valid ADC present:

```
$ auto-discovery --name test --description "some csv" --n_experiments 4 --out_dir ./out data.csv
auto-discovery: error: --model=vertex_ai/gemini-3.7-flash needs Vertex AI configuration. Set VERTEXAI_PROJECT to your Google Cloud project id and VERTEXAI_LOCATION to the Vertex region, usually 'global'.
$ echo $?
1
```

And the prefix error, previously a ~20-line traceback:

```
$ auto-discovery --model gemini-3.7-flash ... data.csv
auto-discovery: error: 'gemini-3.7-flash' is missing a provider. Model names are litellm-qualified as <provider>/<model>, e.g. vertex_ai/gemini-3.7-flash or openai/gemini-3.7-flash.
```

`pytest -m "not modal and not adc"`: 198 passed, plus one pre-existing unrelated failure (`test_run_cell_subprocess_success`, a matplotlib cache-dir permission issue in this sandbox; it fails identically on `main` at `f18a616`). The focused provider test file passes all 42 tests. `ruff check --ignore T201` and `ruff format --check` are clean on the touched files; the two ignored `T201` findings are pre-existing intentional console output.

## One thing for you to decide

**The CHANGELOG entry is under `## Unreleased`, and no version is bumped.** The repo's pattern is a deliberate release step (`make set-version` / `push-version-tag`), and #79 is a sibling fix you may want to ship in the same 1.0.1 — so batching seemed better than bumping here. Say the word and I'll bump it.

Note this is now a rename, not only a fix: any deployment setting the old names must be updated, and anyone whose ADC quota project happened to be their Vertex project was working by accident and will get a startup error. Both are deliberate — the startup error is a named one-line message, which is the outcome #78 asks for.

Suggested-by: @dirkraft


dirkraft:
@gas2own actually, I didn't realize litellm actually introduced its own names for these two vertex vars. If we onyl support those can we eliminate some of our own introduced complexity? Given that the developer already has to set the project, its not much to also have them set location to global (or whatever they prefer) at the same time and eliminate any login or defaulting behavior of our own.

gas2own:
@dirkraft I'm still working on this — the inline turn hit its time limit, so I moved it to the long runtime. I'll post the result here when it finishes.

_ran via Claude_

gas2own:
@dirkraft Yes — done, and the result is a net deletion. Pushed to [#83](https://github.com/allenai/asta-autodiscovery/pull/83) (CI green).

`VERTEX_PROJECT_ID`/`VERTEX_LOCATION` were this package's own names for two settings litellm already reads from the environment, and keeping them is what required the mapping onto `vertex_project`/`vertex_location`, the location default, and a precedence order between two name sets for one setting — all of it upstream of both #78 bugs. The package now names litellm's `VERTEXAI_PROJECT`/`VERTEXAI_LOCATION` and reads them in exactly one place, to check both are set. `_provider_kwargs`, `_vertex_project`, `_vertex_location` and `VERTEX_DEFAULT_LOCATION` are gone; nothing is forwarded, defaulted or inferred.

I kept the startup check, and extended it to cover the location too. It isn't defaulting logic — it's the thing that makes the *absence* of defaulting safe, because each litellm fallback is a silent mid-run 404 rather than an error (project → whatever ADC carries, location → `us-central1`, which doesn't serve the default Gemini models). It's also what makes the rename safe to do loudly: a deployment still setting only the old names now fails at startup naming the new ones instead of drifting onto a fallback.

One thing worth knowing that I confirmed against litellm 1.97's source while doing this: litellm has *two* pairs of names, and they aren't interchangeable. It reads the short `VERTEX_PROJECT`/`VERTEX_LOCATION` only in `embedding()` — `_complete_vertex_ai` and `_complete_vertex_ai_beta` read the `VERTEXAI_` pair alone. So `VERTEXAI_*` is the only configuration every call this package makes can see, and the check rejects the short aliases deliberately rather than by omission. That's recorded next to the tuple that names them, with a test.

The rename also landed in the Cloud Run deploy script, the docker jobs env passthrough, and `.env.example` — that last one was still handing `docker compose` developers the dead names, so a fresh checkout would have hit the new startup error. The deploy script's location default moved from `us-west1` to `global` at the same time; `us-west1` doesn't serve the default models either, so it was the same bug pinned to a different region.

Two things still open from earlier in this PR: the CHANGELOG entry is under `## Unreleased` with no version bumped, on the assumption you'd want to ship this with #79 as one 1.0.1 — say the word and I'll bump. And since the diff changed shape substantially since your review, I can fire an independent AI review over it before you read it again.

_ran via Claude_

packages/autodiscovery/src/autodiscovery/llm.py
```diff
@@ -51,8 +54,23 @@
...
+#:
+#: These names specifically, and not litellm's shorter ``VERTEX_PROJECT`` /
+#: ``VERTEX_LOCATION`` aliases: litellm reads the short pair only in
+#: ``embedding()``, while every completion path reads the ``VERTEXAI_`` pair
+#: alone. Only these two are honoured by every call this package makes.
+_VERTEX_ENV_VAR_HELP = {
+    "VERTEXAI_PROJECT": "your Google Cloud project id",
+    "VERTEXAI_LOCATION": "the Vertex region, usually 'global'",
+}
+VERTEX_ENV_VARS = tuple(_VERTEX_ENV_VAR_HELP)
```
dirkraft: this feels overly complicated. just code against the two vars rather than create this collection @gas2own

gas2own: @dirkraft TL;DR: Addressed in [commit 05e8dc4](https://github.com/allenai/asta-autodiscovery/commit/05e8dc48935a99e0d9b4b7e6d9f932c91eccaa18): removed the collection and now check both env vars directly. All 42 focused tests pass; CI unit/UI/public E2E are green, with Docker and password-file E2E still running. Review re-requested on [PR #83](https://github.com/allenai/asta-autodiscovery/pull/83). No knowledge-base update was needed.

**Decisions needed**

- Should this remain under Unreleased for a combined 1.0.1 with #79, or should I bump the version now?

_ran via Codex_

dirkraft (approved):

dirkraft:
checked out this branch and saw the effects of this change. early exit due to expected VERTEXAI_* not being set. changed my local env from the old to new vars. cli works fine.

PR #84, opened 2026-09-03
Stop run_mcts from spinning when no node can be committed

### LINKED ISSUE #79: run_mcts spins forever when the data-loader node can't be created, writing GBs of log

gas2own:
Fixes #79.

A durable model failure at the data-loader step used to spin `run_mcts` forever
(3.9 GB of log in 90 s, three of those lines the actual error). Two independent
guards, as the issue prescribes:

1. **Bound no-progress iterations in `run_mcts`.** Consecutive iterations that
   commit no node are counted; past `max_no_progress_iterations` (new keyword
   argument, default 3) the run aborts with a new `run.NoProgressError` naming
   the last underlying failure. Whatever was explored is saved first, then the
   error propagates, so the CLI exits non-zero. Running out of selectable nodes
   is still a normal ending — it is only treated as a failure when expansion was
   erroring or when the run never committed a single node.
2. **Stop returning `[root]` unconditionally.** `select_nodes` returns `[]` once
   the root has no untried experiments and cannot generate more, so the existing
   `if not next_nodes: break` escape becomes reachable.

The duplicated per-node failure reporting in the sequential and threaded
expansion paths is folded into one helper, which is also where the last error is
recorded.

### Acceptance

- [x] A run whose experiment generation always fails terminates with a non-zero
      exit and surfaces the underlying error.
- [x] Log output for that run is bounded (at most `max_no_progress_iterations`
      iterations of the repeated lines).
- [x] Regression tests in `packages/autodiscovery/tests/test_run_no_progress.py`
      drive `run_mcts` with an always-failing generator and assert termination,
      the surfaced cause, and the bounded iteration count. A fifth test pins the
      benign case: a tree that runs dry after committing work still exits
      cleanly.

### Verification

Against the pre-fix code, a script reproducing the issue (data-loader expansion
always raising) had to be killed at a 15 s timeout. With the fix it terminates
immediately:

```
TERMINATED: NoProgressError: Exploration aborted: no node is left to expand
after 1 consecutive iteration(s) that committed no node (0/50 experiments
completed). Last failure: RuntimeError: Vertex AI project not set
```

`uv run --all-packages pytest -m "not modal and not adc"` — 192 passed, 1 failed.
The failure is `packages/code_execution/tests/test_ipython_session.py::test_run_cell_subprocess_success`,
which also fails on a clean checkout of `main` in this sandbox (subprocess
IPython), unrelated to this change.

Co-authored-by: @dirkraft

### DIFF
--- a/packages/autodiscovery/src/autodiscovery/mcts_utils.py
+++ b/packages/autodiscovery/src/autodiscovery/mcts_utils.py
@@ -266,11 +266,14 @@ def select_nodes(selection_method, root, nodes_by_level, n_warmstart=0, return_n
         return_n: Number of nodes to return.
 
     Returns:
-        List of selected MCTSNode instances for expansion.
+        List of selected MCTSNode instances for expansion. Empty when nothing is left to expand.
     """
     # If the data loader node hasn't been executed, return the root. This is not run in batch mode.
+    # Once the root can no longer produce that experiment -- its data-loader experiment was consumed
+    # by an expansion that failed and it is not allowed to generate more -- return nothing, so the
+    # caller stops instead of selecting the root forever.
     if len(nodes_by_level[1]) == 0:
-        return [root]
+        return [root] if root.has_untried_experiments() else []
 
     # If there are warmstart experiments left to run, select the data loader node.
     n_children_at_data_loader = len(nodes_by_level[2])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,23 @@ All notable changes to the published packages — [`asta-autodiscovery`][pypi]
 
 [pypi]: https://pypi.org/project/asta-autodiscovery/
 
+## Unreleased
+
+### Fixed: a run whose experiment generation always fails now stops ([#79])
+
+A durable model failure at the data-loader step — misconfigured Vertex
+project/location, expired credentials, exhausted quota, replies that never parse
+into an experiment — used to spin `run_mcts` at full speed forever, emitting
+gigabytes of the same four log lines and burying the underlying error. It now
+aborts, saves whatever was explored, and raises `run.NoProgressError` naming the
+last failure, so the CLI exits non-zero.
+
+`run.run_mcts(...)` takes a new `max_no_progress_iterations` keyword argument
+(default `3`): the number of consecutive iterations that may commit no node
+before the run aborts.
+
+[#79]: https://github.com/allenai/asta-autodiscovery/issues/79
+
 ## 1.0.0
 
 First release of the litellm-based model layer ([#67], [#68]). Every model call
--- a/packages/autodiscovery/src/autodiscovery/run.py
+++ b/packages/autodiscovery/src/autodiscovery/run.py
@@ -282,6 +282,27 @@ def compute_and_store_reward(
         # TODO: Update all past nodes with the new surprisal set
 
 
+class NoProgressError(RuntimeError):
+    """Raised when exploration stops committing nodes and cannot recover.
+
+    Attributes:
+        iterations: Number of consecutive iterations that committed no node.
+        last_error: The last per-node expansion error, when one was raised.
+    """
+
+    def __init__(self, message, *, iterations, last_error=None):
+        """Initialize the error.
+
+        Args:
+            message: Human-readable description of the stall.
+            iterations: Number of consecutive iterations that committed no node.
+            last_error: The last per-node expansion error, when one was raised.
+        """
+        super().__init__(message)
+        self.iterations = iterations
+        self.last_error = last_error
+
+
 def run_mcts(
     root,
     nodes_by_level,
@@ -324,6 +345,7 @@ def run_mcts(
     batch_size=1,
     n_threads=1,
     agent_usage_mode: str = "per_response",
+    max_no_progress_iterations: int = 3,
 ):
     """Run AutoDS exploration. In MCTS, root node level=0 is a dummy node with no experiment, level=1 is the first real node with the dataset loading experiment, levels > 1 are the actual MCTS nodes with hypotheses and experiments.
 
@@ -371,6 +393,13 @@ def run_mcts(
         agent_usage_mode: Tracking mode for agents chat usage. ``per_response`` records
             usage from each AG2 model response. ``summary_delta`` records usage from
             AG2 usage-summary deltas.
+        max_no_progress_iterations: Number of consecutive iterations that may commit no
+            node before exploration aborts. Guards against a durable failure (bad model
+            credentials, exhausted quota, unparseable replies) spinning the loop forever.
+
+    Raises:
+        NoProgressError: If ``max_no_progress_iterations`` consecutive iterations commit
+            no node. Whatever was explored before the abort is still saved.
     """
 
     def _get_executor_rich_outputs(code_executor_agent) -> list:
@@ -405,6 +434,8 @@ def run_mcts(
 
     usage_tracker = UsageTracker()
     usage_tracker.save_events(log_dirname)
+
+    no_progress_error: NoProgressError | None = None
 
     try:
         if agent_usage_mode == "per_response":
@@ -474,8 +505,34 @@ def run_mcts(
         total_to_sample = max_iterations
         n_sampled = 0
         iteration_idx = 0
+        consecutive_no_progress = 0
+        last_expansion_error: Exception | None = None
+        expansion_error_guard = threading.Lock()
         node_mutation_locks: dict[str, threading.Lock] = {}
         node_mutation_locks_guard = threading.Lock()
+
+        def _on_expand_error(node_id: str, exc: Exception) -> None:
+            """Report a failed node expansion and remember it as the latest failure cause."""
+            nonlocal last_expansion_error
+            with expansion_error_guard:
+                last_expansion_error = exc
+            # Keep exploration running when one node expansion fails.
+            print(f"[run_mcts] Failed expanding node {node_id}: {exc.__class__.__name__}: {exc}")
+            traceback.print_exception(type(exc), exc, exc.__traceback__)
+
+        def _no_progress_error(reason: str) -> NoProgressError:
+            """Build the abort error for a stalled run, naming the last underlying failure."""
+            cause = (
+                f"{last_expansion_error.__class__.__name__}: {last_expansion_error}"
+                if last_expansion_error is not None
+                else "no usable experiment was generated"
+            )
+            return NoProgressError(
+                f"Exploration aborted: {reason} "
+                f"({n_sampled}/{total_to_sample} experiments completed). Last failure: {cause}",
+                iterations=consecutive_no_progress,
+                last_error=last_expansion_error,
+            )
 
         def _get_node_mutation_lock(node_obj: MCTSNode) -> threading.Lock:
             """Return a per-node lock used to serialize mutable node state updates."""
@@ -501,6 +558,14 @@ def run_mcts(
                 return_n=min(batch_size, remaining_budget),
             )[:remaining_budget]
             if not next_nodes:
+                # Running out of nodes is the normal way exploration ends early. It is a failure
+                # only when the tree ran dry because expansion kept erroring, or when the run never
+                # committed a single node (e.g. the data loader could never be created).
+                if consecutive_no_progress and (last_expansion_error is not None or n_sampled == 0):
+                    raise _no_progress_error(
+                        f"no node is left to expand after {consecutive_no_progress} "
+                        "consecutive iteration(s) that committed no node"
+                    ) from last_expansion_error
                 break
             print(
                 "SAMPLED "
@@ -772,14 +837,6 @@ def run_mcts(
                         is_threaded=True,
                     )
 
-                def _on_expand_error(node_id: str, exc: Exception) -> None:
-                    # Keep exploration running when one node expansion fails.
-                    print(
-                        f"[run_mcts] Failed expanding node {node_id}: "
-                        f"{exc.__class__.__name__}: {exc}"
-                    )
-                    traceback.print_exception(type(exc), exc, exc.__traceback__)
-
                 expanded_nodes = []
                 with ThreadPoolExecutor(max_workers=n_threads) as executor:
                     future_labels = {
@@ -834,11 +891,7 @@ def run_mcts(
                         )
                     except Exception as exc:
                         # Align sequential behavior with threaded mode by continuing after per-node failures.
-                        print(
-                            f"[run_mcts] Failed expanding node {node_id}: "
-                            f"{exc.__class__.__name__}: {exc}"
-                        )
-                        traceback.print_exception(type(exc), exc, exc.__traceback__)
+                        _on_expand_error(node_id, exc)
                         continue
                     if new_node is not None:
                         expanded_nodes.append(new_node)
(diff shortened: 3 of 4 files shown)


dirkraft:
note to self, I don't like how much execution path complexity this adds into the already beastly run_mcts 600 line function. The general guard this PR attempts to introduce is robust, but the original symptom was me forgetting to set one VERTEX_ var. Going to stew on it a bit.

dirkraft:
@gas2own can you propose changes to run_mcts to reduce execution complexity, particularly so that the anti-spinning changes of this PR land in a way that is easier to understand?

gas2own:
Started async job `gas2own-adfd2ed0c5d8` (2/10 slots in use). I’ll post its result here when it finishes.

gas2own:
@dirkraft Pushed a refactor commit to the PR branch ([`8762d37d`](https://github.com/allenai/asta-autodiscovery/pull/84/commits/8762d37d)) that takes the anti-spin logic *out* of `run_mcts` rather than adding to it. The guard's four mutable locals, its lock and its two closures are now one `ProgressGuard` object in a new `autodiscovery/progress.py`, leaving three self-describing call sites in the loop:

```python
progress.check_exhausted(n_sampled)                      # selection returned nothing
progress.record_expansion_error(node_id, exc)            # per-node failure hook (both paths)
progress.check_iteration(len(expanded_nodes), n_sampled) # end of each iteration
```

Saving the run also moves into the existing `finally`, which deletes the deferred-raise dance entirely — the sentinel local, the extra `except` clause and the trailing re-raise all go, and `NoProgressError` just propagates. Net effect on `run.py` is **+15 lines over `main` instead of +89**, and the guard now has unit tests that don't have to drive a whole exploration.

Two bigger cuts I'd take as separate PRs rather than bolt onto this one — details below, say which you want.

**Decisions needed**

1. Moving the save into `finally` means an *unexpected* exception now also saves the partial run instead of discarding it. I think that's right, but it's a behaviour change beyond issue #79 — happy to revert to the explicit two-path save if you'd rather keep the blast radius at zero.
2. Stage 2/3 below (lifting `_expand_node` out, collapsing the threaded/sequential fork) — in this PR, as follow-ups, or not at all.

An independent review of the updated diff is running and will post to the PR.

_ran via Claude (claude-opus-5, high) · shortened — ask for the full reply_

gas2own:

## P0 (blocking)

None.

## P1 (should fix)

- `packages/autodiscovery/src/autodiscovery/progress.py:69`: Any successful expansion clears `last_error`, so a partially successful final batch can consume failed experiments, exhaust the tree, and exit cleanly while under budget. Track failures per iteration separately and let `check_exhausted` report exhaustion caused by failed expansions.
- `packages/autodiscovery/src/autodiscovery/run.py:858`: Persistence in `finally` can mask the intended `NoProgressError` if `save_nodes` or usage saving fails. Preserve and re-raise the primary exception, reporting persistence failures separately or chaining them without replacing it.
- `packages/autodiscovery/src/autodiscovery/progress.py:39`: `max_no_progress_iterations` is not validated; zero or negative values produce nonsensical threshold behavior. Require a positive integer at construction or `run_mcts` entry.

## P2 (nits)

None.

gas2own:
Resolved the three findings from the [independent review](https://github.com/allenai/asta-autodiscovery/pull/84#issuecomment-5784846348) in commit [`32861f58`](https://github.com/allenai/asta-autodiscovery/commit/32861f58):

1. **Mixed success on exhaustion:** An expansion failure remains visible when another expansion commits in the same iteration. If the tree then runs dry, `check_exhausted` raises `NoProgressError`; a later clean iteration clears a stale error. Added guard regressions for both cases.
2. **Save failure masking the primary error:** A failed `save_nodes` or usage save is reported to stderr without replacing an active exploration exception; when no primary error exists, the save failure still propagates. Added failing-node-save and failing-usage-save regressions.
3. **Invalid no-progress limit:** `ProgressGuard` rejects zero, negative, non-integer and boolean limits. Added parameterized validation tests.

Validation: focused tests 20 passed; credential-free suite 206 passed with two unrelated tests excluded. The unfiltered suite reports failures in `test_models_litellm_has_not_mapped_still_resolve` (model now in LiteLLM catalog) and `test_run_cell_subprocess_success` (subprocess IPython in this environment). Ruff formatting passes; Ruff lint flags existing issues across the touched modules. No review findings remain open; the updated PR is being independently re-reviewed.

gas2own:
Started async job `gas2own-305cd89de758` (2/10 slots in use). I’ll post its result here when it finishes.

gas2own:
The [independent review](https://github.com/allenai/asta-autodiscovery/pull/84#issuecomment-5784846348) found three should-fix issues and no blocking P0 issue. I addressed all three in [commit `32861f58`](https://github.com/allenai/asta-autodiscovery/commit/32861f58) and posted a [finding-by-finding resolution](https://github.com/allenai/asta-autodiscovery/pull/84#issuecomment-5784930836). A second independent review is running; the human reviewer is already requested. Local tests: 206 passed, with two unrelated failures excluded.

**Decisions needed**
- Keep saving partial runs after unexpected exceptions, or restore the prior behavior?
- Pursue the larger `run_mcts` simplifications as follow-up PRs, in this PR, or not at all?

_ran via Codex (gpt-6-sol, high)_

gas2own:

## P0 (blocking)
None.

## P1 (should fix)
- `packages/autodiscovery/src/autodiscovery/run.py:859`: The use of `sys.exception()` requires Python 3.11+. If this project supports Python 3.10 or older, this will raise an `AttributeError` during cleanup, masking the original error and skipping the remaining `finally` block statements. Replace it with `sys.exc_info()[1]` which is compatible with all Python 3 versions.
- `packages/autodiscovery/src/autodiscovery/run.py:876`: The 2-argument form of `traceback.print_exception(persistence_error, file=sys.stderr)` requires Python 3.10+. Older versions require the legacy 3-argument form. Given that `progress.py:70` uses the legacy 3-argument form, this introduces a runtime crash on Python 3.9 if supported. Use `traceback.print_exception(type(persistence_error), persistence_error, persistence_error.__traceback__, file=sys.stderr)` for consistency and compatibility.

## P2 (nits)
- `packages/autodiscovery/src/autodiscovery/progress.py:70`: If the project actually requires Python 3.11+ (making the `sys.exception()` usage safe), you can simplify the legacy `traceback.print_exception(type(exc), exc, exc.__traceback__)` to just `traceback.print_exception(exc)` to match the modern Python idiom used in `run.py`.

Recommendation: approve-with-nits

gas2own:
The [second AI review](https://github.com/allenai/asta-autodiscovery/pull/84#issuecomment-5785069710) found no blocking issue that applies to this PR: both compatibility warnings concern older Python versions, while the workspace and package require Python ≥3.13. The remaining suggestion is stylistic, so I made no change and did not request another automatic review. ([raw.githubusercontent.com](https://raw.githubusercontent.com/allenai/asta-autodiscovery/32861f58091673afb66ddb56079a0db54979f51f/pyproject.toml))

The human reviewer is already requested. The PR currently conflicts with `main` and will need reconciliation before merging.

**Decisions needed**
- Keep saving partial runs after unexpected exceptions, or restore the previous behavior?
- Put the larger `run_mcts` simplifications in this PR, pursue them as follow-ups, or leave them out?

_ran via Codex (gpt-6-sol, high)_

packages/autodiscovery/src/autodiscovery/mcts_utils.py
```diff
@@ -266,11 +266,14 @@ def select_nodes(selection_method, root, nodes_by_level, n_warmstart=0, return_n
...
-        List of selected MCTSNode instances for expansion.
+        List of selected MCTSNode instances for expansion. Empty when nothing is left to expand.
     """
     # If the data loader node hasn't been executed, return the root. This is not run in batch mode.
+    # Once the root can no longer produce that experiment -- its data-loader experiment was consumed
+    # by an expansion that failed and it is not allowed to generate more -- return nothing, so the
+    # caller stops instead of selecting the root forever.
     if len(nodes_by_level[1]) == 0:
-        return [root]
+        return [root] if root.has_untried_experiments() else []
```
dirkraft: @gas2own curious. if this was the only line we changed, would that guard against the spinning i experienced by simply forgetting to set a VERTEX env var?

gas2own: @dirkraft Yes. For the missing-`VERTEX_*` case, this one line on its own stops the spin. After one failed iteration the loop exits, and the error is printed once.

What happens in that run, going by the code on `main`:

1. The root starts with exactly one queued experiment (the data loader) and `allow_generate_experiments=False`.
2. In iteration 1, `select_nodes` returns `[root]`. `get_next_experiment` takes the data-loader experiment off the queue, which needs no model call. The child node's chat then fails on Vertex. The child is rolled back, and the traceback is printed once. Those are the few real error lines in your log.
3. From iteration 2 on, on `main`: `nodes_by_level[1]` is still empty, so `select_nodes` returns `[root]` again. The root now has nothing queued and can't generate more, so `_expand_node` prints "No new experiment generated… Skipping" and returns `None`. `n_sampled` never moves, so this repeats forever. That loop is the 3.9 GB, and it involves no model calls at all.
4. With this line, `root.has_untried_experiments()` is `False` in iteration 2. `select_nodes` returns `[]`, and the existing `if not next_nodes: break` ends the loop.

Two things this line alone does not give you:

- **The run exits 0.** After the `break`, `run_mcts` saves an empty tree and returns normally. The CLI reports success with 0 of N experiments done, and the cause is only visible in the traceback printed back in iteration 1.
- **A failure after the data loader still spins.** Only the root has this "nothing left to try" exit. Take a model or config failure that only appears deeper in the tree, for example in code execution or belief scoring. Each iteration there pops or generates an experiment, fails, and rolls back. The failure goes through the exception path, which never sets `allow_generate_experiments=False`, so the node stays selectable and `n_sampled` never advances. That is still an endless loop, just a slower one, and it makes a model call every iteration. The no-progress counter is aimed at this case.

So if you want the smallest possible change: this line fixes the exact thing you hit. Adding a non-zero exit when the loop ends with nothing committed would also make the failure show up at the CLI. The counter is only needed for the deeper-tree case.

_ran via Claude (claude-opus-5-5, high) · shortened — ask for the full reply_

dirkraft: @gas2own please reduce the robustness of this PR to this bare minimum

gas2own: @dirkraft I cut PR 84 down to the one-line fix in [`5f1e1b23`](https://github.com/allenai/asta-autodiscovery/commit/5f1e1b23). The diff against `main` now touches three files: the `select_nodes` guard in `mcts_utils.py`, one 26-line test in `test_select_nodes.py`, and a short CHANGELOG entry.

- **Removed:** the no-progress counter, `NoProgressError`, the `progress.py` module with its tests, and every change to `run.py`.
- **Merge conflict:** I merged `main` into the branch and resolved the CHANGELOG conflict, so the PR can merge again.
- **Checks:** the new test passes and Ruff formatting and lint are clean on the touched files. I did not re-run the full test suite.
- **What this fix doesn't cover:** the PR description now says so. A run that fails at the data loader still exits 0. A lasting failure deeper in the tree can still loop.
- **Updated:** the title and description to match. I didn't start another automatic review for a one-line change.

The two earlier open questions no longer apply, because the code they were about is gone: whether to save partial runs from `finally`, and the larger `run_mcts` refactor.

_ran via Claude (claude-opus-5-5, high)_

packages/autodiscovery/src/autodiscovery/mcts_utils.py
```diff
@@ -266,11 +266,13 @@ def select_nodes(selection_method, root, nodes_by_level, n_warmstart=0, return_n
...
     Returns:
-        List of selected MCTSNode instances for expansion.
+        List of selected MCTSNode instances for expansion. Empty when nothing is left to expand.
     """
     # If the data loader node hasn't been executed, return the root. This is not run in batch mode.
+    # Once the root has nothing left to try (its data-loader expansion failed), return nothing so
+    # the caller stops instead of selecting the root forever.
     if len(nodes_by_level[1]) == 0:
-        return [root]
+        return [root] if root.has_untried_experiments() else []
```
dirkraft: @radamson is there any obvious issue with this change?

radamson (approved):

packages/autodiscovery/src/autodiscovery/mcts_utils.py
radamson: No obvious issue (better to stop than repeatedly fail), but would like to monitor it to see how often this occurs...

PR #86, opened 2026-09-03
Replace the hand-maintained LLM price table with litellm's cost data

### LINKED ISSUE #70: Replace the hand-maintained LLM price table with litellm's cost data
Split out of #68 (per review) so that PR stays scoped to provider/model routing. Related: #67.

## Problem

`api/metrics/costs.py` carries a hand-maintained price table and a provider-prefix stripper:

```python
# Pricing per 1M tokens (USD) — Vertex AI standard pricing
LLM_PRICING: dict[str, dict[str, float]] = {
    "gemini-3.1-pro-preview": {"input": 2.00, "output": 12.00},
    ...   # 26 entries
}

# Fallback pricing for unknown models (conservative: uses gemini-3.1-pro rates)
_FALLBACK_PRICING = {"input": 2.00, "output": 12.00}


def _lookup_pricing(model_name: str) -> dict[str, float]:
    if model_name in LLM_PRICING:
        return LLM_PRICING[model_name]
    # Strip provider prefix (e.g., "google/gemini-3.1-pro-preview" -> "gemini-3.1-pro-preview")
    stripped = model_name.split("/")[-1] if "/" in model_name else model_name
    ...
    return _FALLBACK_PRICING
```

This is the last piece of model-name-specific logic in the repo after #68 — same category as the `is_gemini_model` / `google/` prefix handling that #67 removed everywhere else. It has three concrete failure modes:

1. **Unknown models are silently mispriced.** `_FALLBACK_PRICING` is Gemini 3.1 Pro's rate ($2/$12 per 1M). Any model not in the 26-entry table — including every model added after the table was last touched — is billed at Gemini Pro rates with no warning in the output.

2. **Prefix stripping discards the provider.** `github_copilot/gpt-4o` and `openai/gpt-4o` both strip to `gpt-4o` and get OpenAI list price. Copilot doesn't bill per token at all (it bills "premium requests"), so any Copilot run's reported cost is meaningless. #68 makes mixed-provider runs expressible, so this is now reachable rather than theoretical.

3. **It drifts by construction.** Prices are a snapshot with no update mechanism, and the header says "Vertex AI standard pricing" while the table also contains OpenAI models.

## Why this is worth doing now

After #68, litellm is the transport for every model call, and it **already computes the cost of each call** — `response._hidden_params["response_cost"]`, derived from its own maintained per-model `input_cost_per_token` / `output_cost_per_token`.

We currently throw that away. `LiteLLMAG2Client.get_usage()` surfaces it to AG2, but `UsageTracker` does not: `_extract_usage_from_response` reads only `model` and the token counts, and `_empty_bucket()` has no cost field. So the authoritative number is computed on every call and discarded, and then approximated later from a stale table.

## Proposal

**Record the cost at the source, not in the dashboard.**

1. Have `UsageTracker.record_response` read `_hidden_params["response_cost"]` and store it on the event; add `cost_usd` to `_empty_bucket()` so it aggregates into `by_model` / `by_agent` / `by_node` / `by_component` alongside the token counts.
2. `api/metrics/costs.py` then sums a field instead of pricing anything. `calculate_llm_cost` keeps its signature and return type, so `aggregator.py` (3 call sites: lines 191, 398, 952) is unaffected.
3. Delete `LLM_PRICING`, `_FALLBACK_PRICING` and `_lookup_pricing`.

Note that `_lookup_pricing` is imported directly by `aggregator.py:29` and used at line 191, so that call site needs looking at too — it prices a single model outside `calculate_llm_cost`.

**Explicitly not** adding litellm to `api/`. It is a separate lightweight Flask service with its own `requirements.txt`, pinned to Python 3.11, and pulling litellm in for a price lookup is a bad trade. Keeping the cost computation on the producing side avoids that entirely.

## Backwards compatibility

Runs completed before this change have `llm_usage_summary.json` files with no cost field. The dashboard needs to either:

- fall back to the current table for those runs (keeping it around as a legacy path, defeating much of the point), or
- report cost as unavailable for pre-change runs, or
- backfill the summaries.

Worth a decision before implementing. Reporting "unavailable" is probably right — the old numbers are approximations anyway, and silently mixing two costing methods in one dashboard is worse than a gap.

## Also worth checking

- Whether litellm has cost data for `github_copilot/*` at all. If it does not, the honest output is "cost unavailable for this provider" rather than a token-based estimate — see failure mode 2.
- `calculate_llm_cost` currently derives output tokens as `total - prompt` to capture reasoning tokens. litellm's own cost already accounts for reasoning tokens, so that arithmetic goes away with the table.

gas2own:
Closes #70.

`api/metrics/costs.py` carried a 16-entry price table, a provider-prefix stripper and a Gemini-3.1-Pro fallback rate. litellm is already the transport for every model call and already computes each call's cost from its own maintained prices — we were discarding that and approximating it later from a stale snapshot. This records it at the source instead, and the dashboard now sums a field.

## What changed

**Producer** (`packages/autodiscovery`)

- `llm.reported_cost(response)` — the single reader of `_hidden_params["response_cost"]`.
- `llm.cost_of(response, model)` — returns `{total_usd, prompt_usd, completion_usd, reasoning_usd}`. `response_cost` is the authoritative total and is never recomputed; the split comes from `litellm.cost_per_token` and is *scaled onto* that total, because `response_cost` also covers cache reads, image input and built-in tools that a plain prompt/completion split does not.
- `UsageTracker.record_event(..., cost=)` stores it per event; `record_response(..., request_model=)` prices the call. Pricing needs the provider and a response only reports a bare model name, hence the extra argument — wired through all five call sites including the sandbox image-analysis path, which prices in-sandbox and carries the result out through its usage marker.
- `_empty_bucket()` gains `cost_usd`, `prompt_cost_usd`, `completion_cost_usd`, `reasoning_cost_usd` and `calls_with_cost`, so cost aggregates into `by_model` / `by_agent` / `by_node` / `by_component` alongside the token counts.

**Consumer** (`api/`)

- `LLM_PRICING`, `_FALLBACK_PRICING`, `_lookup_pricing` and `_lookup_bucket_cost_by_type` are gone. `calculate_llm_cost` keeps its signature and return type. The `aggregator.py:191` call site you flagged (`_lookup_pricing` imported directly) went with it — `_build_event_cost_breakdowns` now sums recorded event costs rather than pricing each bucket.
- litellm is **not** added to `api/requirements.txt`.

## The three questions the issue left open

**Backwards compatibility → report unavailable.** As you suggested. Pre-change runs have no cost field, so `cost_status()` reads `unavailable` and the dashboard shows a dash with a tooltip instead of a dollar figure; aggregate cards carry a quiet `excludes N runs with no recorded cost` caveat. No legacy pricing path is kept.

**Does litellm have cost data for `github_copilot/*`? No.** All 33 of its `github_copilot` registry entries carry a null input and output cost — consistent with Copilot billing premium requests rather than tokens. litellm surfaces that as a *zero* price, not an absent one, so a naive "use `response_cost`" would have reported every Copilot run at exactly $0.00 — a different wrong answer than before, and a harder one to notice. `cost_of` therefore treats a zero cost over a non-empty prompt as unpriced and returns `None`. That distinction is the reason for `calls_with_cost` rather than a plain cost field: it is what separates "this call was free" from "nobody knows what this call cost".

**`total - prompt` output-token arithmetic → gone**, as you expected. Reasoning tokens are a subset of `completion_tokens` in the usage shape litellm normalizes to, so `reasoning_usd` is now carved *out of* the completion cost. The old dashboard code added them (`explicit_output_tokens = completion_tokens + reasoning_tokens`), which double-counted.

## Cost status, three states

`llm_cost_status` is `complete` (every call priced), `partial` (some priced — total renders as `≥ $x.yz`) or `unavailable`. `partial` is reachable in one run: a mixed-provider run after #68, where the Vertex calls price and the Copilot ones do not.

`AggregatedUsageBucket` gains `priced_calls` alongside `total_calls`. One recording path — AG2's `summary_delta` mode — knows a call's total but not its split, so the stacked cost bars fall back to a single neutral segment rather than rendering as empty.

## Not done here

`by_model` is still keyed by the bare model name the response reports, so `openai/gpt-4o` and `github_copilot/gpt-4o` still share a row in that one breakdown. Cost is unaffected — it is summed per call, not priced per label — so this is now cosmetic rather than a mispricing. Qualifying the key would split every existing model row in two across the old/new boundary, which seemed worth doing deliberately rather than as a side effect of this change.

## Testing

30 new tests (`test_llm_cost.py`, `test_llm_usage.py`, `api/tests/test_costs.py`, `api/tests/test_aggregator_costs.py`) covering: Copilot and unmapped models pricing to `None`; the split summing to the injected `response_cost` including when the two disagree; reasoning carved out of completion; and a pre-change usage summary reading as `unavailable`. `make test` is 218 passed, 1 failed — the failure is `test_ipython_session.py::test_run_cell_subprocess_success`, which cannot write `$HOME/.cache/matplotlib` in this sandbox and touches no cost code. `tsc --noEmit` and `eslint` are clean on the UI; `next build` was OOM-killed in the sandbox, so a full UI build is unverified here and CI should be watched.

Suggested-by: @dirkraft

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,30 @@ All notable changes to the published packages — [`asta-autodiscovery`][pypi]
 
 [pypi]: https://pypi.org/project/asta-autodiscovery/
 
+## Unreleased
+
+### Runs record what each model call cost ([#70])
+
+A run's `llm_usage_summary.json` and `llm_usage_events.jsonl` now carry the cost
+litellm computed for each call, alongside the token counts. Every usage bucket
+gains `cost_usd`, `prompt_cost_usd`, `completion_cost_usd`, `reasoning_cost_usd`
+and `calls_with_cost`; each event gains a `cost` object.
+
+Cost is reported as **unavailable** — not as zero — where it is genuinely not
+known. litellm prices every `github_copilot` model at zero because Copilot bills
+"premium requests" rather than tokens, so no honest per-token figure exists for
+it, and models litellm has not mapped yet have no price at all.
+`calls_with_cost` is what tells an unpriced call apart from a free one.
+
+This replaces the metrics dashboard's hand-maintained price table, which billed
+any unlisted model at Gemini 3.1 Pro's rate and priced `github_copilot/gpt-4o`
+as OpenAI's `gpt-4o`. Runs that completed before this change carry no cost field
+and the dashboard reports their cost as unavailable; their previous figures were
+approximations from that table, and a gap is better than silently mixing two
+costing methods in one dashboard.
+
+[#70]: https://github.com/allenai/asta-autodiscovery/issues/70
+
 ## 1.0.0
 
 First release of the litellm-based model layer ([#67], [#68]). Every model call
--- a/api/metrics/aggregator.py
+++ b/api/metrics/aggregator.py
@@ -26,7 +26,16 @@ from typing import Any
 from autodiscovery_jobs.config import JobConfig
 from google.cloud import storage
 
-from .costs import _lookup_pricing, calculate_llm_cost, get_duration_seconds
+from .costs import (
+    UNAVAILABLE,
+    bucket_call_counts,
+    bucket_costs,
+    calculate_llm_cost,
+    coerce_float,
+    coerce_int,
+    cost_status,
+    get_duration_seconds,
+)
 from .models import (
     AggregatedUsageBucket,
     AggregatedUsageResponse,
@@ -73,6 +82,7 @@ class JobSnapshot:
     model: str | None = None
     llm_usage_summary: dict | None = None
     llm_cost_usd: float = 0.0
+    llm_cost_status: str = UNAVAILABLE
     llm_cost_by_agent: dict[str, dict[str, float]] = field(default_factory=dict)
     llm_cost_by_component: dict[str, dict[str, float]] = field(default_factory=dict)
     llm_cost_by_node: dict[str, dict[str, float]] = field(default_factory=dict)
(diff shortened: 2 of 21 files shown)


dirkraft:
abandoning for the moment. I did not expect this many LoC. Not a priority.

PR #89, opened 2026-09-08
Cap code-executor output so a printed dataset can't OOM the job

gas2own:
The prod AutoDiscovery Cloud Run job was SIGKILLed (`exit(137)`) twice on 2026-09-07, ~5 minutes apart. Both executions died on the turn right after a generated code block returned an enormous stdout — one logged `[CodeExecutor] Stdout length: 675997343 characters` (676 MB) after the agent read a raw `.sql` dump and printed its table lines. This is not rare: executions on 2026-09-08 logged 160 MB and 82 MB outputs from the same pattern.

`ModalSandboxExecutor.execute_code_blocks` returned `result.stdout` uncapped. That string becomes a `CodeResult.output`, which becomes a group-chat message, which is then re-serialized into every subsequent LLM request — so one 676 MB print is copied many times over and the container is killed on the next speaker.

This caps the complete code result at a fixed 20,000 characters, keeping the head and tail (headers/schema at the front, final result or traceback at the back) and dropping the middle. Twenty thousand characters is roughly 5,000 LLM tokens and leaves about 10,000 characters for each end before the elision notice—enough for many tabular rows or a typical traceback without letting one tool result dominate future model requests. The elision notice counts toward the cap, states how many characters were omitted, and tells the agent to aggregate, sample, or write to a file instead.

Tests in `packages/autodiscovery/tests/test_code_output_truncation.py` cover the exact boundary, head/tail retention, the fixed total ceiling across stdout and stderr, and a 5 MB stdout going through `execute_code_blocks`.

Not addressed here: the giant string is still materialized once inside the executor before the cap applies, so a single output large enough to exhaust memory on its own would still fail. Capping at the sandbox boundary (`asta-sandbox`) would close that; this change removes the amplification, which is what actually killed the job.

Related: #79 (unbounded log volume from a spinning `run_mcts`) is a separate symptom of the same missing-output-bounds theme.

Suggested-by: @dirkraft

### DIFF
--- a/packages/autodiscovery/src/autodiscovery/agents.py
+++ b/packages/autodiscovery/src/autodiscovery/agents.py
@@ -40,28 +40,39 @@ IMAGE_ANALYST_PROMPT = """Please analyze the given plot image and provide the fo
 # becomes a chat message that is then re-serialized into every subsequent LLM
 # request — so an uncapped output is copied many times over and OOM-kills the job.
 # Observed in production: a single block emitted 676 MB of stdout and the container
-# was SIGKILLed (exit 137) on the next turn.
-DEFAULT_MAX_CODE_OUTPUT_CHARS = 200_000
+# was SIGKILLed (exit 137) on the next turn. 20k characters is roughly 5k LLM
+# tokens: enough for about 10k characters from each end (including a typical
+# traceback or many tabular rows) without letting one tool result consume a large
+# fraction of a future model request.
+MAX_CODE_OUTPUT_CHARS = 20_000
 
 
-def _truncate_output(output: str, max_chars: int) -> str:
-    """Clip an over-long code output to ``max_chars``, keeping both ends.
+def _truncate_output(output: str) -> str:
+    """Clip an over-long code output to the fixed cap, keeping both ends.
 
     The head carries whatever the code printed first (headers, schema, counts) and
     the tail carries the final result, so both are worth keeping; only the middle
-    is dropped. A non-positive ``max_chars`` disables truncation.
+    is dropped. The truncation notice itself counts toward the cap.
     """
-    if max_chars <= 0 or len(output) <= max_chars:
+    if len(output) <= MAX_CODE_OUTPUT_CHARS:
         return output
 
-    dropped = len(output) - max_chars
-    notice = (
-        f"\n\n... [output truncated: {dropped} of {len(output)} characters omitted; "
-        f"limit is {max_chars}. Do not print whole datasets — aggregate, sample, or "
-        f"write to a file and read back only what you need.] ...\n\n"
-    )
-    head = max_chars // 2
-    tail = max_chars - head
+    dropped = len(output) - MAX_CODE_OUTPUT_CHARS
+    while True:
+        notice = (
+            f"\n\n... [output truncated: {dropped} of {len(output)} characters omitted; "
+            f"limit is {MAX_CODE_OUTPUT_CHARS}. Do not print whole datasets — "
+            f"aggregate, sample, or write to a file and read back only what you "
+            f"need.] ...\n\n"
+        )
+        retained = MAX_CODE_OUTPUT_CHARS - len(notice)
+        actual_dropped = len(output) - retained
+        if actual_dropped == dropped:
+            break
+        dropped = actual_dropped
+
+    head = retained // 2
+    tail = retained - head
     return output[:head] + notice + output[-tail:]
 
 
@@ -121,7 +132,6 @@ class ModalSandboxExecutor(CodeExecutor):
         vision_model: str,
         timeout: int = 30 * 60,
         usage_tracker: UsageTracker | None = None,
-        max_output_chars: int | None = None,
     ):
         """Initialize the sandbox executor wrapper.
 
@@ -130,19 +140,9 @@ class ModalSandboxExecutor(CodeExecutor):
             timeout: Timeout in seconds (for Autogen compatibility)
             vision_model: Vision model, as litellm's ``<provider>/<model>``
             usage_tracker: Optional usage tracker for image analysis calls.
-            max_output_chars: Cap on the output handed back to the chat, overriding
-                ``AUTODISCOVERY_MAX_CODE_OUTPUT_CHARS`` and
-                ``DEFAULT_MAX_CODE_OUTPUT_CHARS``. Non-positive disables the cap.
         """
         self._executor = backend
         self._timeout = timeout
-        if max_output_chars is None:
-            max_output_chars = int(
-                os.environ.get(
-                    "AUTODISCOVERY_MAX_CODE_OUTPUT_CHARS", DEFAULT_MAX_CODE_OUTPUT_CHARS
-                )
-            )
-        self._max_output_chars = max_output_chars
         self.vision_model = vision_model
         self._usage_tracker = usage_tracker
         self._usage_node_id: str | None = None
@@ -215,13 +215,13 @@ class ModalSandboxExecutor(CodeExecutor):
 
             print(f"[CodeExecutor] Stdout length: {len(output)} characters")
 
-            output = _truncate_output(output, self._max_output_chars)
+            output = _truncate_output(output)
             if len(output) != len(result.stdout or ""):
                 print(f"[CodeExecutor] Stdout truncated to {len(output)} characters")
 
             if result.stderr:
                 print(f"[CodeExecutor] Stderr: {result.stderr[:200]}")
-                stderr = _truncate_output(result.stderr, self._max_output_chars)
+                stderr = _truncate_output(result.stderr)
                 output += f"\nSTDERR:\n{stderr}"
 
             if not result.success:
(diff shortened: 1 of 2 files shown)


packages/autodiscovery/src/autodiscovery/agents.py
```diff
@@ -35,6 +35,35 @@
 4. Annotations and Legends: Describe key annotations or legends.
 5. Statistical Insights: Provide insights based on the information presented in the plot."""
 
+# Cap on how much of a code block's output is kept. Generated code regularly prints
+# an entire dataset (a raw .sql dump, a full dataframe), and the executor's output
+# becomes a chat message that is then re-serialized into every subsequent LLM
+# request — so an uncapped output is copied many times over and OOM-kills the job.
+# Observed in production: a single block emitted 676 MB of stdout and the container
+# was SIGKILLed (exit 137) on the next turn.
+DEFAULT_MAX_CODE_OUTPUT_CHARS = 200_000
```
dirkraft: 1. 200k is still quite a lot for output, especially if this gets included in future llm calls? can you propose a more limited limit with justification or illustration of why that limit makes sense?

packages/autodiscovery/src/autodiscovery/agents.py
```diff
@@ -100,9 +130,19 @@ def __init__(
...
         """
         self._executor = backend
         self._timeout = timeout
+        if max_output_chars is None:
+            max_output_chars = int(
+                os.environ.get(
+                    "AUTODISCOVERY_MAX_CODE_OUTPUT_CHARS", DEFAULT_MAX_CODE_OUTPUT_CHARS
+                )
+            )
+        self._max_output_chars = max_output_chars
```
dirkraft: don't make this configurable, just rely on one hardcoded constant

dirkraft (changes requested):
@gas2own see inline comments

gas2own:
Started async job `gas2own-1bf3038fb1aa` (1/10 slots in use). I’ll post its result here when it finishes.

gas2own:
@dirkraft TL;DR: Addressed both comments in [PR #89](https://github.com/allenai/asta-autodiscovery/pull/89): the cap is now a fixed 20,000 characters (~5,000 tokens), with no configuration overrides. Targeted tests and CI unit/security checks pass; remaining CI and an independent review are running. No knowledge-base update was needed.

_ran via Codex_

gas2own:

## P0 (blocking)

None.

## P1 (should fix)

- `packages/autodiscovery/src/autodiscovery/agents.py:218`: stdout is truncated before stderr/errors are appended, then the combined output is truncated again at line 270. The second notice reports omissions from the already-truncated string, hiding the potentially millions of characters originally dropped. Assemble all sections first and truncate once, or preserve cumulative truncation metadata.

## P2 (nits)

None.

gas2own:
Resolved the independent-review P1 in commit 2ce2d63.

- The executor now assembles stdout, stderr, errors, and image analysis as logical sections and truncates them once, so the notice reports the original combined character count.
- The section-aware truncation avoids materializing a second giant combined string and still retains the overall head and tail.
- Regression coverage now verifies the original 1,000,012-character combined size is reported.

Validation: all 6 output-truncation tests pass; formatting passes. The broader credential-free suite reached 215 passes with one unrelated local subprocess timeout; the same unit suite was green in PR CI before this commit, and fresh CI is running.

Open findings: none.

gas2own:
Started async job `gas2own-e72f7bd0eaed` (1/10 slots in use). I’ll post its result here when it finishes.

gas2own:
## 🤖 Independent AI code review (Claude)

Automated review of the PR by a separately invoked Claude reviewer. Prioritised P0/P1/P2; treat as advisory input for the human reviewer, not a merge gate.

## P0 (blocking)

None.

## P1 (should fix)

- `packages/autodiscovery/src/autodiscovery/agents.py:63-76` (`_truncate_output_parts`, retained/head/tail computation): `retained` can be ≤ 0 if `MAX_CODE_OUTPUT_CHARS` is ever lowered below the notice length (~260 chars). With `retained == 0`, `part[-suffix_remaining:]` becomes `part[-0:]` and returns the *entire* part; with `retained < 0`, `part[:negative]` returns nearly the whole part. The function then silently emits an output far larger than the cap — the exact failure it exists to prevent. Fix: `retained = max(0, MAX_CODE_OUTPUT_CHARS - len(notice))` and skip the slice loops entirely when `head`/`tail` are 0 (`if not suffix_remaining: suffix = ""`), plus assert/clamp the constant to be larger than the notice.
- `packages/autodiscovery/src/autodiscovery/agents.py:244` : `output_parts.append(f"\nSTDERR:\n{result.stderr}")` materializes a second full copy of a potentially 100s-of-MB stderr, which directly contradicts the docstring claim that working from sections "avoids first concatenating potentially enormous stdout and stderr". Fix: append two parts — `output_parts.extend(("\nSTDERR:\n", result.stderr))`.
- `packages/autodiscovery/src/autodiscovery/agents.py:288-292`: truncation happens only *after* the full 676 MB result has been received and held in the executor process. If the OOM/SIGKILL occurs while receiving or holding the sandbox result (not only on the next LLM turn), this change does not fix it. Consider capping at read time in the sandbox backend (bounded read of stdout/stderr) in addition to this post-hoc clip, or at minimum drop the reference (`result.stdout = None` / del) before returning.

## P2 (nits)

- `packages/autodiscovery/src/autodiscovery/agents.py:78-86`: the head/tail split can slice mid-line and can delete the `STDERR:` marker (it lands in the dropped middle when stdout is huge), so the model sees stderr text with no indication of its origin. Consider clipping each section independently (stdout, stderr, error, image analyses) with per-section budgets so the section markers always survive.
- `packages/autodiscovery/src/autodiscovery/agents.py:57-68`: the `while True` fixpoint over the notice's digit count is more machinery than needed; formatting the notice with a fixed-width count (e.g. `{dropped:>12}`) or budgeting `len(str(output_length))` digits up front makes it a single pass and removes the (currently benign) risk of the loop being made non-monotonic by a future edit to the notice text.
- `packages/autodiscovery/src/autodiscovery/agents.py:45`: `MAX_CODE_OUTPUT_CHARS` is hardcoded; make it a `ModalSandboxExecutor.__init__` parameter (defaulting to the constant) so callers/tests can vary it without monkeypatching a module global.
- `packages/autodiscovery/tests/test_code_output_truncation.py:88`: no test for the near-boundary case `output_length == MAX_CODE_OUTPUT_CHARS + 1`, where the notice is longer than the dropped region and the fixpoint loop iterates. Add one asserting `len(result) == MAX_CODE_OUTPUT_CHARS` and that the reported omitted count matches `output_length - (len(prefix)+len(suffix))`.
- `packages/autodiscovery/tests/test_code_output_truncation.py:52,62,71,80`: `from autogen.coding import CodeBlock` / `from asta_sandbox import ExecutionResult` are repeated inside test bodies; hoist to module level.

Recommendation: approve-with-nits

gas2own:
Resolved the latest independent-review findings:

- Tiny-cap invariant: fixed in 602a2ae. `_truncate_output_parts` now clamps retained output at zero and returns a bounded notice instead of allowing Python's `[-0:]` slice to retain a full section. Regression tests cover a tiny cap and the one-character-over-limit boundary.
- Large stderr copy: fixed in 64f657c by keeping the marker and stderr as separate parts.
- Read-time sandbox cap: not changed here. The PR description already identifies sandbox-boundary truncation as separate follow-up work; this PR removes the repeated downstream amplification while retaining the sandbox result once during execution.
- Per-section budgets/marker preservation: not changed. The current behavior intentionally retains the overall output's head and tail; changing allocation among stdout, stderr, errors, and image analysis is a separate output-policy decision.
- Fixpoint simplification: not changed; the loop converges on the exact notice size, and the new near-boundary regression verifies the omitted count.
- Configurable cap: not changed because the requested design is one fixed hardcoded limit.
- Repeated test imports: left as a non-functional style nit.

Validation: all 8 focused output-truncation tests pass; the changed test file passes Ruff and both changed files pass Ruff formatting. The credential-free suite reached 217 passes with only the pre-existing two-second subprocess-startup timeout failing, including when rerun alone.

Open finding: sandbox-boundary/read-time truncation remains outside this PR's scope as documented in the PR description.

gas2own:
TL;DR: The [review](https://github.com/allenai/asta-autodiscovery/pull/89#issuecomment-5608988816) found no P0s. I fixed the remaining actionable issue in `602a2ae`, documented every disposition in the [resolution comment](https://github.com/allenai/asta-autodiscovery/pull/89#issuecomment-5609079370), and all CI checks pass. The sandbox read-time cap remains explicitly out of scope; no knowledge-base update was needed.

**Decisions needed**

- Re-review [PR #89](https://github.com/allenai/asta-autodiscovery/pull/89).

_ran via Codex_

gas2own:
TL;DR: The [blocking review finding](https://github.com/allenai/asta-autodiscovery/pull/89#issuecomment-5608877962) is fixed. A [second review](https://github.com/allenai/asta-autodiscovery/pull/89#issuecomment-5608988816) recommended approval with nits; actionable items were resolved and [documented](https://github.com/allenai/asta-autodiscovery/pull/89#issuecomment-5609079370). All required checks are green, and @dirkraft has been re-requested.

_ran via Codex_

dirkraft (approved):

PR #90, opened 2026-09-08
Add AGENTS.md: public-repo disclosure guardrails for AI agents

gas2own:
## What

Adds `AGENTS.md` — a short guide for AI coding agents working in this repository, built around one thing the repo currently states nowhere: **this repo is public, so treat everything written here as published.**

## Why

Agents routinely arrive in this repo carrying context from somewhere non-public — an internal chat thread, an internal ticket, another repository that is private or internal — and then write a commit message, issue, or PR body that reflects that context. Public artifacts are world-readable the moment they are pushed and stay readable after a delete (forks, caches, the events API), so the boundary can only be enforced *before* the push. Today it depends entirely on the individual agent noticing on its own.

`AGENTS.md` makes it explicit:

- a plain statement that the repo is public and what that means in practice;
- a do-not-write list — secrets, non-public links/contents, internal infrastructure detail, personal data (including real user data as test fixtures), unannounced plans;
- how to credit a human contributor without leaking the non-public thread the guidance came from (verified GitHub handle only);
- the local `make sync` / `make test` / `make lint` expectation before pushing, and branch-and-PR / no-self-merge conventions.

Nothing here changes code or CI — it is documentation that agents load as instructions.

## Note on `CLAUDE.md`

`.gitignore` intentionally treats `CLAUDE.md` as a local file (line 31), so this PR does **not** add one; `AGENTS.md` is the committed source of truth. Contributors whose harness only reads `CLAUDE.md` can point a local, ignored one at it with a single line: `@AGENTS.md`.

## Review focus

The do-not-write list is deliberately conservative. If any item is too broad for how this repo actually works — for example if some infrastructure detail is already public in `docs/` — say so and I will narrow it.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,71 @@
+# Agent guide — asta-autodiscovery
+
+Instructions for AI coding agents working in this repository. Human contributors
+should read [`README.md`](README.md) and [`docs/`](docs/index.md) first.
+
+## This repository is PUBLIC — treat everything you write here as published
+
+`allenai/asta-autodiscovery` is a public repository. Code, commit messages, issue
+and pull-request titles and bodies, review comments, branch names, test fixtures,
+and docs are all world-readable and indexed the moment they are pushed — and they
+stay readable after a delete, via forks, caches, and the events API. There is no
+"take it back."
+
+Agents commonly reach this repo carrying context from somewhere non-public: an
+internal chat thread, an internal ticket, an incident channel, another repository
+that is private or internal. **Before you write anything here, strip the context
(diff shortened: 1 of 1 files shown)


dirkraft (approved):

pauls-ai2:
This just tripped me up today. Looks like we need to have a `CLAUDE.md` file locally with `@AGENTS.md` or else people will get assigned to PRs (and any other rules we add) when Claude creates a PR. Makes me think we may want to consider a `make setup` or something to set some of that stuff up for future developers.

PR #91, opened 2026-09-10
Read metrics cache configuration from the environment

gas2own:
## Problem

The metrics dashboard renders zeros for every metric on any deployment whose GCS bucket is not the built-in default.

`MetricsCache.__init__` built its configuration with `JobConfig()` instead of `JobConfig.from_env()`, so it ignored `GCS_BUCKET` / `AUTODISCOVERY_BUCKET` and scanned the default bucket name rather than the configured one. Every other consumer of `JobConfig` in `api/` threads a `from_env()`-derived config down from the request handler (`JobManager.config`), which is why only the metrics dashboard was affected.

Two behaviours then hid the misconfiguration behind plausible-looking zeros instead of surfacing it:

- `_scan_all_jobs` caught a failed bucket listing and returned an empty `AggregatedData` stamped with a fresh `refreshed_at`. That is indistinguishable from "this deployment has no runs", so the cache reported itself as healthy and current.
- `_do_refresh` stored that empty result, overwriting a previously good scan.

## Changes

- `MetricsCache` reads its configuration via `JobConfig.from_env()`.
- A discovery failure in `_scan_all_jobs` now propagates instead of being swallowed into an empty result.
- `_do_refresh` preserves the last good data when a refresh raises and records a sanitized `refresh_failed` status; exception details remain in logs.
- Automatic refresh attempts observe the normal refresh interval after a failure, avoiding request-driven retry storms; the explicit refresh endpoint still bypasses this backoff.
- `MetricsCache.last_error` is exposed on `GET /metrics/cache/status`, so an unreadable bucket reads as an error rather than as an empty dataset.

## Testing

`api/tests/test_metrics_aggregator.py` covers environment-based configuration, discovery-failure propagation, previous-data preservation, sanitized error status, retry backoff, forced-refresh bypass, and clearing the error on success.

- `uv run --all-packages pytest api/tests/test_metrics_aggregator.py`: 7 passed
- Main suite: 224 passed with one unrelated subprocess test deselected because it fails under the available Python 3.13 runtime
- Ruff checks for the test file and fatal Python errors in the changed files: passed

No configuration change is needed to deploy this: the fix makes the cache honour the bucket the rest of the API already uses.

Suggested-by: @radamson

### DIFF
--- a/api/metrics/aggregator.py
+++ b/api/metrics/aggregator.py
@@ -654,8 +654,16 @@ class MetricsCache:
                 return
             self._last_refresh_attempt_monotonic = now
             self._refreshing = True
-        thread = threading.Thread(target=self._background_refresh, daemon=True)
-        thread.start()
+        try:
+            thread = threading.Thread(target=self._background_refresh, daemon=True)
+            thread.start()
+        except Exception:
+            # A thread that never started cannot run ``_background_refresh``'s
+            # cleanup. Restore the state before preserving the existing error
+            # behavior for the caller.
+            with self._lock:
+                self._refreshing = False
+            raise
 
     def _background_refresh(self) -> None:
         """Run a refresh, but only if no other worker on this pod is already scanning.
--- a/api/metrics/aggregator.py
+++ b/api/metrics/aggregator.py
@@ -548,6 +548,7 @@ class MetricsCache:
         self._refresh_interval = refresh_interval_seconds
         self._refreshing = False
         self._last_error: str | None = None
+        self._last_refresh_attempt_monotonic: float | None = None
         # Must be ``from_env()``: the plain constructor yields the built-in
         # defaults (including the default bucket name), so a deployment that
         # points GCS_BUCKET elsewhere would be scanned against the wrong bucket.
@@ -560,10 +561,11 @@ class MetricsCache:
 
     @property
     def last_error(self) -> str | None:
-        """Message from the most recent failed refresh, cleared on success.
+        """Status code for the most recent failed refresh, cleared on success.
 
         Surfaced by the cache-status endpoint so an unreadable bucket reads as a
-        failure rather than as a legitimately empty dataset.
+        failure rather than as a legitimately empty dataset. Detailed exception
+        text is retained only in logs.
         """
         with self._lock:
             return self._last_error
@@ -627,7 +629,7 @@ class MetricsCache:
 
     def force_refresh(self) -> None:
         """Force a background refresh regardless of staleness."""
-        self._trigger_background_refresh()
+        self._trigger_background_refresh(force=True)
 
     def _is_stale(self) -> bool:
         if not self._data or not self._data.refreshed_at:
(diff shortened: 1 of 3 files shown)


gas2own:

## P0 (blocking)

None.

## P1 (should fix)

- `api/metrics/aggregator.py:699` — Preserved stale data remains stale, so every request can start another refresh after a failure, hammering an unavailable bucket. Track the last attempt and apply the normal interval/backoff before retrying.
- `api/metrics/metrics_api.py:84` — Returning raw exception text can expose bucket names and provider/internal details. Expose a sanitized error code/message; retain full exceptions only in logs.

## P2 (nits)

None.

gas2own:
Addressed both review findings in commit `8feed1c`:

- Retry pressure after a failed refresh: automatic refresh attempts now record their monotonic start time and observe the configured refresh interval before retrying. Explicit refresh requests still bypass the backoff.
- Raw exception disclosure: the cache-status API now reports the stable `refresh_failed` code; full exception details remain only in application logs.

Validation: all 7 metrics-cache tests pass, including new backoff and forced-refresh coverage. The broader suite passes 224 tests with the unrelated Python 3.13 subprocess test deselected; focused Ruff checks and `git diff --check` pass. Nothing remains open from this review.

gas2own:

## P0 (blocking)
- `api/metrics/metrics_api.py:82`: Since `_scan_all_jobs` now raises an exception on failure instead of returning a fallback empty object, if the very first scan fails (when the cache is uninitialized), `cache._data` will remain `None`. The API endpoint unconditionally accesses `data.scan_duration_seconds` (and likely `data.jobs` for `unique_users`), which will cause an `AttributeError` 500 crash rather than returning the intended `last_error` degradation state. Fix: Update `metrics_api.py` to gracefully handle `data is None` (e.g., returning zeros or nulls alongside `cache.last_error`).

## P1 (should fix)
- `api/metrics/aggregator.py:654`: If the background thread fails to launch (e.g., `thread.start()` raises `RuntimeError: can't start new thread`), the `self._refreshing` flag will remain `True` permanently. This will deadlock all future refresh attempts. Fix: Wrap `thread.start()` in a `try...except Exception` block and reset `self._refreshing = False` under the lock in the except block before re-raising or logging.

## P2 (nits)
None.

gas2own:
Resolved the Gemini review against commit `3349a8d`:

- Cold-start `None` crash: not applicable. `get_data()` synchronously assigns `_warm_start()` before launching the background refresh, and `_warm_start()` always returns `AggregatedData`, including when no persisted cache can be loaded. Added regression coverage for the empty cold-start path.
- Failed thread launch: fixed. `_trigger_background_refresh()` now clears `_refreshing` if thread construction or startup raises, while preserving the exception for the caller. Added regression coverage for the failed-start path.

Validation: all 9 metrics-cache tests pass; the broader suite passes 226 tests with the unrelated subprocess test deselected; focused Ruff checks and `git diff --check` pass. Nothing remains open from this review.

dirkraft:
@radamson do you need this? since its only changes to api.metrics.* i'm minimally concerned about any collateral effects to other autodiscovery components so you can move forward as quickly as you'd like. i'm not planning to look at this in any particular level of detail

dirkraft:
bodhi asked about this. it's better on this branch than in main.

main
Screenshot 2026-09-14 at 4 13 36 PM

this branch
Screenshot 2026-09-14 at 4 13 30 PM

dirkraft (approved):

radamson:
Thanks @dirkraft 🙂

PR #94, opened 2026-09-15
Add one-time credit grant script for existing users

pauls-ai2:
Adds `api/scripts/grant_credits.py`, a one-shot maintainer script that adds 1,000 credits to each eligible user. Follows #93, which raised the default.

**Scope.** Only users with an explicit `granted_credits` override. Everyone else falls through to the default that #93 already raised. So the script only ever adds to a value that is already stored, and never needs to know what the default is.

**Re-running is safe.** It appends `{"id": "grant-1000-credits-v1", "amount": 1000, ...}` to a `grants` list on the profile and skips users already carrying that id. The credit and the record of it are one write, so an interrupted run resumes by re-running.

**Concurrent writes.** Profiles are read with their GCS generation and written back with `if_generation_match`. A profile that changed underneath is reported as a conflict and left alone.

**`UserProfile` is unchanged.** The script reads and writes `user.json` directly. The diff to that module is 18 lines of docstring recording that `grants` exists and that `update_user_profile` drops it.

### Checks

- `make test`: 250 pass. One unrelated failure, `test_provider_args.py::test_models_litellm_has_not_mapped_still_resolve`, fails identically on `main`.
- 12 new tests covering eligibility, the appended record, re-run skipping, conditional writes and both conflict paths. `ruff` clean.
- **Not yet run against a real bucket** — verified against mocks only. Please `--dry-run`, with `--userid` on one account first, before a real pass.


packages/autodiscovery_jobs/src/autodiscovery_jobs/user_profile.py
```diff
@@ -30,6 +30,13 @@ class UserProfile:
         granted_credits: Custom credit allocation for user (None = use default)
         created_at: ISO timestamp when profile was created
         updated_at: ISO timestamp when profile was last updated
+
+    ``from_dict`` keeps only the attributes above, so any other key a stored
+    user.json holds is dropped when the profile is saved through this class.
```
dirkraft, line 35: > any other key a stored
    user.json holds is dropped when the profile is saved through this class

this is ok?

dirkraft: maybe a different question: why remark about this at all? are we intentionally deprecating some old fields which may still be present in existing data?

pauls-ai2: I guess it isn't needed. It's just kinda how python works.

pauls-ai2: The thing it was talking about, without saying it; There are props in the json object which are not in the class (namely a new `grants` array, which tracks what one-time grants have been applied).

dirkraft: Right, so those will be lost if an instance of UserProfile is serialized and persisted back into gcs?

PR #98, opened 2026-09-17
Keep the run experiments list payload bounded

### LINKED ISSUE #97: Run experiments endpoint returns ~80 MiB payloads, exceeding Cloud Run's 32 MiB response limit → persistent 500s on large runs
## Summary

`POST /api/runs/{userid}/{runid}/experiments` returns the **full experiment tree including every node's `code` and `code_output`**. For a large run this reaches ~80 MiB of JSON, which gzips to just over Cloud Run's **32 MiB HTTP/1 response limit** — so `prod-proxy` returns **500** to the browser even though nginx and the Flask app both logged `200`.

Because the UI only records an experiment id as "known" *after* a successful response, a 500 means it never advances: the 15-second poll re-requests the entire tree forever. One run generated 42 identical 500s in ~2 hours, and the response grew (79.8 MiB → 82.5 MiB) as the job wrote new nodes.

## Evidence

Cloud Run request log (`prod-proxy`) and the app's own access log for the same request:

| layer | status | size |
|---|---|---|
| Flask / nginx `response_length` | 200 | 86,490,419 B (82.5 MiB) |
| nginx `body_bytes_sent` (gzipped) | 200 | 33,914,729 B (32.3 MiB) |
| Cloud Run request log | **500** | — |

Every 5xx in the window had `body_bytes_sent > 32 MiB`; every response under 32 MiB on the same endpoint returned 200 (e.g. 0.85 MiB and 1.2 MiB polls succeeded). Latency was a consistent ~4.2–4.7 s, so this is not the 60 s proxy timeout from #75.

## Root cause

1. **`api/utils/experiments.py` — `ExperimentNode.to_dict()`** includes `code` and `code_output` for every node. This is the list-payload serializer, so the whole tree's captured stdout ships on every poll.

   The detail route looks like it was written expecting the opposite: `api/runs/runs_api.py` (`get_run_experiment_details`) does `experiment_node["code_output"] = node.code_output` *after* calling `to_dict()` — a no-op today, which suggests `code_output` was meant to be detail-only and leaked into the list payload.

2. **`ui/src/app/contexts/RunExperimentsContext.tsx`** adds ids to `knownExperimentIds` only inside the success path, so the `known_experiment_ids` incremental-fetch optimisation can never engage once the response exceeds the limit. The failure is self-sustaining rather than transient.

## Suggested fix

- **Drop `code_output` (and likely `code`) from the list payload** in `ExperimentNode.to_dict()`, keeping them on `GET /experiments/{experiment_id}`, which already fetches per-node detail. In the UI, `codeOutput` is consumed only by `ExperimentDetails` (which uses the detail fetch) and `exportUtils`; the export path would need to hydrate from the detail endpoint, and `ExperimentGraph` already tolerates `codeOutput: null`.
- **Guard the response size server-side** so the endpoint degrades instead of 500ing — e.g. cap the number of nodes returned per response and let the client page through with `known_experiment_ids`, which is exactly what that parameter is for.
- **Make the client resilient**: on a failed poll, fall back to a smaller request (or record ids from a partial/paged response) so a single oversized payload can't wedge the run page permanently.
- Optionally add a log-based alert on `response_length` over ~25 MiB to catch this before it becomes a 5xx.

## Impact

The run detail page is unusable for large runs — the experiments table never loads, and the retry loop keeps the `autodiscovery_5xx_count` alert firing.

gas2own:
Fixes #97.

`POST /api/runs/{userid}/{runid}/experiments` serialized the whole experiment
tree including every node's `code` and `code_output` — unbounded captured source
and stdout. On a large run the response outgrew the serving layer's
response-size limit and the browser got a 5xx, even though the app itself logged
a 200.

## Changes

**API**

- `ExperimentNode.to_dict()` omits `code`/`code_output` unless
  `include_code=True`. Only the per-experiment detail route passes it — which is
  what that route already looked like it intended, since it reassigned
  `code_output` after calling `to_dict()` (a no-op). That line is gone.

**UI**

- `code`/`code_output` are typed as optional on the list payload.
- The JSON export hydrates `code`/`codeOutput` from the detail route in bounded
  batches (6 at a time); a missing or failed detail response aborts the export
  with the existing error toast rather than silently writing an incomplete file.
  The Download button shows "Preparing…" while that runs. The CSV export never
  used either field. `ExperimentDetails` already reads from the detail fetch and
  `ExperimentGraph` already tolerates a null `codeOutput`.

## Deliberately not handled

Per review feedback, this PR no longer paginates the list response. An earlier
revision added a cursor, a per-page cap, and a multi-page drain in the UI
polling loop; all of that is reverted. The request body is back to
`known_experiment_ids` and the response has no `has_more`/`next_cursor`.

If a list response still exceeds the serving layer's limit for some other reason
— a very large number of experiments, or very large per-node `analysis` /
`review` / `experiment_plan` text — it is left unhandled for now; the extra code
was judged not worth it. The log-based alert on `response_length` suggested in
the issue is a monitoring-config change, so it is also out of scope here.

## Validation

- `uv run --all-packages pytest api/tests/test_experiments.py` → 3 passed,
  covering the detail-only code fields on both `to_dict()` paths and the list
  serializer.
- Ruff: no new findings on the changed Python files relative to `main` (these
  files carry pre-existing findings).
- `yarn ts-check` / `yarn lint` locally, with the PR CI run as the full gate.

Reviewer requested: `pauls-ai2`

Suggested-by: @dirkraft


gas2own:
Started async job `gas2own-f1f22cc69b04` (1/10 slots in use). I’ll post its result here when it finishes.

gas2own:

## P0 (blocking)

None.

## P1 (should fix)

- `ui/src/app/contexts/RunExperimentsContext.tsx:270` — A completed run with more than 50 pages stops polling while `has_more` is still true, permanently omitting remaining experiments. Track the final `hasMore` value and only stop polling when completion is true and no pages remain.
- `api/runs/runs_api.py:1149` — Pagination sends every previously received ID back on every request; sufficiently large runs will eventually exceed request-body limits, recreating the failure this change addresses. Use a server-issued cursor or stable offset instead of an ever-growing ID list.
- `ui/src/app/utils/exportUtils.ts:178` — Detail-fetch failures are swallowed, producing a successful but silently incomplete JSON export. Collect failures and reject the export (or explicitly notify the user and require confirmation).
- `ui/src/app/utils/exportUtils.ts:166` — Hydration only occurs when both fields are null, so an experiment missing exactly one of `code` or `codeOutput` remains incomplete. Change the condition to `exp.code == null || exp.codeOutput == null`.

## P2 (nits)

None.

gas2own:
Addressed all four findings in f13d6e7:

- 50-page completion gap: polling now stops only when the job is complete **and** the final fetched page reports no remainder. A completed run above the per-poll drain cap continues on the next interval.
- Growing request body: replaced `known_experiment_ids` with a fixed-size numeric cursor returned by the server and retained by the UI.
- Silent partial JSON exports: missing or failed detail responses now reject the export, which surfaces through the existing download error toast instead of creating an incomplete file.
- Partial hydration: experiments are hydrated when either `code` or `codeOutput` is missing.

Also fixed the four `no-void` errors from the initial UI Build run.

Validation: API pagination tests (6 passed), `yarn ts-check`, and `yarn lint`. The full Python suite reached 243 passed / 1 skipped with one executor-specific subprocess test failure outside this diff; the local Next.js build was killed by the executor memory limit, so the fresh PR CI run will provide the build result.

Open findings: none.

gas2own:
Started async job `gas2own-c7d2192684e7` (1/10 slots in use). I’ll post its result here when it finishes.

gas2own:
The [review](https://github.com/allenai/asta-autodiscovery/pull/98#issuecomment-5718475247) found four P1s and no P0s; all are fixed in [PR #98](https://github.com/allenai/asta-autodiscovery/pull/98), with [resolutions documented](https://github.com/allenai/asta-autodiscovery/pull/98#issuecomment-5718624007) and fresh CI fully green. A follow-up independent review is running, so the PR remains draft. No knowledge-base update was needed.

_ran via Codex_

gas2own:

## P0 (blocking)
- `api/utils/experiments.py:118`: `ExperimentNode.to_dict()` now entirely omits the `"code"` and `"code_output"` keys when `include_code` is false. In `api/runs/runs_api.py:1149`, `ExperimentModel(**node)` will crash with a Pydantic `ValidationError` if the underlying model schema for these fields lacks an explicit default (e.g., `= None`). Fix: Explicitly assign `None` instead of conditionally omitting the keys (e.g., `node["code"] = self.code if include_code else None`) to satisfy strict model validation.

## P1 (should fix)
- `ui/src/app/utils/exportUtils.ts:161`: `hydrateExperimentsForExport` introduces an N+1 API request pattern by fetching the missing `code` and `codeOutput` fields individually for every experiment. For large runs, this will hammer the backend with requests, lock the user in a prolonged "Preparing..." state, and abort the entire export if even a single request fails due to transient network issues. Consider moving this logic to a dedicated bulk-export endpoint on the backend.

## P2 (nits)
- `api/runs/models.py:312`: The description for the `experiments` field in `GetRunExperimentsResponseModel` still reads "excluding the ids the client already has". Update this text to reflect the new cursor-based pagination logic.

gas2own:
Follow-up review resolution for [the Gemini findings](https://github.com/allenai/asta-autodiscovery/pull/98#issuecomment-5718692207):

- Omitted `code` / `code_output` keys: no change. `ExperimentModel` defines both fields with explicit `None` defaults, so constructing it from the list serializer does not raise a validation error. The targeted API tests exercise this path and pass (6/6).
- Per-experiment export hydration: retained as an explicit design tradeoff. It runs only for JSON export, limits concurrency to six requests, and fails the export rather than silently dropping data. A bulk response containing every experiment's unbounded source and output would recreate the response-size risk this PR fixes; a server-side streaming/file export would be a separate product design, not a correctness fix for this PR.
- Stale response-field description: fixed in commit `4edda40`; it now describes cursor-based paging.

Validation: `uv run --all-packages pytest api/tests/test_experiments.py` (6 passed); `git diff --check` passed. The repository-wide Ruff check continues to report pre-existing findings in `api/runs/models.py`, and the fresh CI run is the authoritative full gate.

Open blocking findings: none. The automatic cycle did not reach two-backend agreement on one executable-code head; this last change is documentation-only, so no further automatic review was fired. Human review can decide whether the separate server-side export design is worth follow-up.

gas2own:
Independent check of this PR while working #97. One new finding on the cursor change, plus verification of the two open items from the latest review.

## New finding — the numeric offset cursor can permanently skip an experiment

`api/runs/runs_api.py` pages with `offset=req.cursor` into the `creation_idx`-sorted
node list and returns `next_cursor = cursor + len(page)`. That is only safe if the
list is append-only. It is not: **node files appear out of `creation_idx` order by
default**, because expansion is parallel and a node's JSON is written only after its
experiment finishes.

- `packages/autodiscovery/src/autodiscovery/mcts.py:40` — `creation_idx` is assigned
  at construction from a global counter.
- `packages/autodiscovery/src/autodiscovery/run.py:710` and `:767` — when
  `n_threads > 1 and len(next_nodes) > 1`, nodes are expanded on a `ThreadPoolExecutor`.
- `packages/autodiscovery/src/autodiscovery/run.py:690` — the file is written in
  `_backprop_and_save()` → `save_mcts_node(...)`, i.e. after the experiment completes.
- `packages/autodiscovery/src/autodiscovery/args.py:100-111` — `--batch_size` and
  `--n_threads` both default to **2**, and
  `packages/autodiscovery_jobs/src/autodiscovery_jobs/backends/base.py:91-99` passes
  neither, so hosted runs get the parallel path.

So if nodes 11 and 12 expand concurrently and 12 finishes first:

1. Poll sees `[... 10, 12]`, client advances the cursor past position of 12.
2. Node 11's file lands. The sorted list becomes `[... 10, 11, 12]`, and every
   index from 11 onward shifts by one.
3. The next request at the old cursor returns node 12 again (deduplicated by id
   client-side) — and node 11 is never served.

Node 11 stays missing from the experiments table and the graph for the rest of the
session, and `has_more` gives no hint, since `next_cursor < len(tree)` is satisfied by
the shifted tail. Id-based exclusion (`known_experiment_ids`) did not have this
failure mode, because it is insertion-order independent; it was traded away for the
bounded request body, which is a real concern, but the replacement is not equivalent.

Options, cheapest first:

- Return the tree's total node count and let the client re-drain from cursor `0`
  whenever the number of experiments it holds is below that total after a drain. The
  request body stays one integer, the overlap is deduplicated by id already, and a
  late arrival is picked up on the next poll.
- Treat the cursor as the client's highest *contiguous* `creation_idx` rather than a
  raw offset, so a lower-numbered late arrival is still ahead of the watermark.
- Keep id-based exclusion and bound the *response* by serialized bytes instead of a
  node count, which removes the request-body growth concern's severity without
  relying on list stability.

## The two open items from the Gemini review

- **P0 (`ExperimentModel(**node)` raises when the keys are omitted) — does not
  reproduce.** `code` and `code_output` are declared `str | None = Field(None, ...)`
  in `api/runs/models.py`, so they are optional with a default. Constructing
  `ExperimentModel(**node.to_dict())` against this PR's head succeeds and yields
  `code=None, code_output=None`; the detail path with `include_code=True` yields the
  real values. No change is needed for that finding.
- **P2 (stale field description) — valid.** `GetRunExperimentsResponseModel.experiments`
  still reads "excluding the ids the client already has", which describes the previous
  id-based scheme rather than the cursor. `to_experiment_models`'
  `exclude_experiment_ids` parameter also no longer has a caller.

Separately, one behaviour worth a look regardless of the above: the fixed
`EXPERIMENT_PAGE_SIZE = 200` bounds node *count*, not response size. With `code` and
`code_output` gone the remaining per-node fields are much smaller, so 200 nodes is
comfortably within limits today — but `analysis`, `review`, and `experiment_plan` are
still unbounded text, so the bound is indirect.

gas2own:
The [follow-up review](https://github.com/allenai/asta-autodiscovery/pull/98#issuecomment-5718692207) had no valid blocker: the Pydantic concern was a false positive, while the stale cursor description is fixed in `4edda40`. [Resolutions are documented](https://github.com/allenai/asta-autodiscovery/pull/98#issuecomment-5718738438); [PR #98](https://github.com/allenai/asta-autodiscovery/pull/98) is ready, fully green, and awaiting Paul’s review. No knowledge-base update was needed.

**Decisions needed**

- Human review should decide whether bounded per-item JSON hydration is acceptable or warrants a separate server-side export design. The automatic cycle ended without same-head two-backend agreement.

_ran via Codex_

dirkraft:
@gas2own i'd rather not introduce pagination yet. Just the changes to exclude outputs from the list response is enough. If responses still go over 32MB for whatever reason (huge list of experiments/metadata conceivably, or individual experiment has too much output) I don't want to do anything special for now; not worth the LoC. Please reduce the scope of this PR.

gas2own:
@dirkraft Scope reduced and pushed to `fix/experiments-list-payload-size` ([fb2e705](https://github.com/allenai/asta-autodiscovery/pull/98/commits/fb2e70586fd42903872844fa742b6ff6159b6264)) — the cursor pagination is gone. The request body is back to `known_experiment_ids`, the response has no `has_more`/`next_cursor`/page cap, and `RunExperimentsContext` is back to its original single-request polling. What's left is the `include_code` flag on `to_dict()` (only the detail route passes it), the now-optional `code`/`code_output` typing on the list payload, and the dropped no-op `code_output` reassignment. Diff went from +319/−56 to +164/−17, and the PR body now records that an over-limit response from anything else is deliberately unhandled. CI is running on the new head.

One judgement call I made rather than guessing silently: I kept the JSON export's per-experiment hydration of `code`/`codeOutput` (~40 lines, rewritten as simple bounded batches instead of the old worker pool). Without it the Download → JSON file silently loses both fields, since the export serializes whatever the list poll loaded. Say the word if you'd rather accept that regression and drop it too — that's a one-commit removal.

**Decisions needed**

- Keep or drop the JSON-export hydration (drop = JSON export no longer contains `code`/`code_output`).

_ran via Claude · shortened — ask for the full reply_

dirkraft:
@gas2own you've got a CI failure

gas2own:
@dirkraft CI is green again on [7cd9ffe](https://github.com/allenai/asta-autodiscovery/pull/98/commits/7cd9ffe). The failure was the UI Build lint step only: after the scope reduction, the `.filter(...)` line in `hydrateExperimentsForExport` was 103 columns, and `prettier/prettier` wanted it wrapped. One-line formatting fix, no behavior change. All seven checks now pass (UI Build, Unit tests, Docker Compose Build, both E2E jobs, GitGuardian).

PR is ready and still awaiting `pauls-ai2`'s review.

_ran via Claude · shortened — ask for the full reply_

dirkraft (approved):

